PR #7, opened 2019-08-30
chore: use caret version range in `composer.json`

nunomaduro:
This pull request makes the file `composer.json` a little bit more clear using `Caret Version Range` by default.


freekmurze:
Thanks!

PR #11, opened 2019-08-30
chore: adds missing extension

nunomaduro:
The class `Facade\Ignition\Views\Compilers\BladeSourceMapCompiler` is using `mb_substr`, therefore ignition needs `ext-mbstring` extension.


freekmurze:
Thanks!

PR #15, opened 2019-08-30
Update the readme to fix #14 and update the badges

### LINKED ISSUE #14: The badges link to the wrong packages
I was going through the readme, and I noticed they're linking to `facade/flare` instead of `facade/ignition` (just thought I'd make a note of it here)

Also, I'm really digging the UI of the package. It looks quite solid

Cool debug package

austinkregel:
At the moment the scrutinizer and the circleci badges 404 because those projects don't seem to be set up. 

I'm leaving them there and making the assumption that the wonderful folks who own this project will set those up. If there are no plans to set up either of those projects I'll be happy to remove the badges from this PR.


freekmurze:
Thank you!

PR #17, opened 2019-08-30
Use .env variables in the config

cca-bheath:
Since this could be dev specific choices on a project or even a workstation, let's make it a bit more flexible.


freekmurze:
Makes sense, thanks!

PR #19, opened 2019-08-30
Added a new configuration setting allowing specified request parameters to be masked

aran112000:
Added a new configuration setting allowing specified request parameters to be masked so they're not exposed (useful for things like passwords and credit card details).

### DIFF
--- a/config/ignition.php
+++ b/config/ignition.php
@@ -24,4 +24,20 @@ return [
     */
 
     'theme' => 'light',
+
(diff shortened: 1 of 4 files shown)


freekmurze:
Could you resolve the conflict with the master branch and add some tests to prove that the new code works?

aran112000:
Hi @freekmurze, sure thing - will push up the tests in a moment

Issue #20, opened 2019-08-30
Not working in versions < 5.7


conroyp:
* Laravel Version: 5.5.48, 5.6.39

### Description:
In the documentation, support is mentioned for versions >= 5.5:

```
In Laravel 5.5, 5.6, 5.7 and 5.8 you can pull in the package via Composer:
```

After adding the package and setting the handler, in versions < 5.7 when an error is generated, the attempt to generate the `ShareReportController` url fails. This causes the old-style whoops page to still be generated.

The error comes from `Illuminate\Routing\UrlGenerator@formatAction`, and is:
```
strpos() expects parameter 1 to be string, array given
```

This is initially triggered by https://github.com/facade/ignition/blob/master/src/ErrorPage/ErrorPageViewModel.php#L99
```
protected function shareEndpoint()
{
    return action([ShareReportController::class]);
}
```

It looks like it's relying on code introduced by https://github.com/laravel/framework/pull/24738, but it seems that was only introduced at 5.7 (confusion about differing versions in documentation has been cleared up previously - https://github.com/laravel/framework/issues/24770)

Updating L99 in `ErrorPageViewModel` to the nastier-looking `return action(['\Facade\Ignition\Http\Controllers\ShareReportController']);` does allow the share url (`/flare/share-report`) to be generated when the changes from https://github.com/laravel/framework/pull/24738 are manually applied, so it seems related to that specific PR.

### Steps To Reproduce:
1. Inside a fresh Laravel project with either `5.5*` or `5.6.*` set as version numbers, and run `composer require --dev facade/ignition`

2. Paste the following code into `routes/web.php`:
```
Route::get('/error', function() {
    I_AM_BROKEN_CODE
});
```

3. Visit `http://{YOUR_PROJECT_URL}/error`

freekmurze:
This is fixed via #21

PR #21, opened 2019-08-30
Add support for L5.5 & 5.6

conroyp:
The [docs](https://flareapp.io/docs/ignition-for-laravel/installation) currently list support for Laravel versions 5.5 and 5.6, but because of a dependency on callable array syntax functionality in the `action` function, any error triggered on a 5.5 or 5.6 app falls back to the old-style `whoops` screen (https://github.com/facade/ignition/issues/20).

This PR changes the `action` call into the `ShareReportController` to the uglier, older style until such a time as support for 5.5 and 5.6 is officially dropped.

There's a similar `action()` call in this file also relying on the callable syntax, for telescope logging. That has been left alone here - it doesn't need to change as telescope itself has a minimum version of 5.7, where this problem no longer exists.

### DIFF
--- a/src/ErrorPage/ErrorPageViewModel.php
+++ b/src/ErrorPage/ErrorPageViewModel.php
@@ -96,7 +96,11 @@ class ErrorPageViewModel implements Arrayable
 
     protected function shareEndpoint()
     {
-        return action([ShareReportController::class]);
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks!

PR #23, opened 2019-08-31
Added a new configuration setting allowing specified request parameters to be masked

aran112000:
Added a new configuration setting allowing specified request parameters to be masked so they're not exposed (useful for things like passwords and credit card details).


jetwes:
Shouldn't this happen in your models itself ($hidden)?

aran112000:
Hi @jetwes,

I could be mistaken, but I don't see how that would affect any GET / POST data that's being printed in plain text on the error pages, and sent across to Flare if set up?

Do you perhaps have an example so I could confirm?

aran112000:
Hi @freekmurze,

Am I right in thinking you're still setting up Travis CI, seems it's not only this build failing due to Xdebug / PHPUnit version differences to doing a composer install outside of Travis CI as it's loading different versions by the looks of things?

freekmurze:
@aran112000 yeah, I still need to finish setting up Travis. I'll do that next week. If there's anything that you spot now that could fix the tests on Travis, feel free to send a PR 👍

jetwes:
@aran112000 Ok - got it. I thought you meant the info displayed in the users tab for example. Your suggestion totally make sense.

freekmurze:
Thank you for this, but we're going to pass on this. We'll provide a better why to achieve this later.

aran112000:
Hi @freekmurze,

Could you perhaps elaborate on this? It's going to be essential for us to use in production as request data will contain all sorts from credit card details being posted to login credentials on the registration/login controllers.

Many thanks

dwightwatson:
Any update on this please? It's quite important that sensitive information isn't being shipped off elsewhere. A good starting point/default would be `Illuminate\Exceptions\Handler::$dontFlash` which will filter out `password` and `password_confirmation` by default.

PR #34, opened 2019-09-01
Add link to Awesome list

mikebronner:
HI guys, I started a repo to collect a list of all igniter plugins. As I was making this for myself, I though I might as well make it into one of those "Awesome" lists and share it with the community. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,6 @@ Please see [CONTRIBUTING](CONTRIBUTING.md) for details.
 ## License
 
 The MIT License (MIT). Please see [License File](LICENSE.md) for more information.
+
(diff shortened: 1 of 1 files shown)


freekmurze:
Hi Mike, thank you for your proposal. In the coming days we are going to add a list of third party Ignition plugins in our own docs. 

Currently people can't send a PR to our docs, but we'll make some change to enable that in the next few days.

mikebronner:
@freekmurze No problem, sounds good!

Gummibeer:
I think that composer suggestions will be too much over time!? And filter them would be unfair.

PR #37, opened 2019-09-01
Format fixes and some improvements for file parsing

DonJorgen:
- Change the way we get parse the files to only return those that we are interested in by passing a filter.

Signed-off-by: Jorge Y. C. Rodriguez <self@jorgerodriguez.dk>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 composer.lock
 .php_cs.cache
 .phpunit.result.cache
+/.idea/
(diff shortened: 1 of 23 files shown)


freekmurze:
Thank you for this.

Because there are so many formatting changes, it’s hard to see if you made any functional ones.

Could you split this in two PRs? One with the functional changes and another one with the changes to the styleci config file alone (no need to fix all the styling issues after that, StyleCI will take care of it)

DonJorgen:
@freekmurze Will do that :)

PR #38, opened 2019-09-01
Automatically fix blade variable typos and optional variables

MrRio:
# This PR adds a solution for fixing undefined variables in view files.

Screen Shot 2019-09-01 at 14 34 03

## The Solution Provider

- It starts by looking at how similar the variable name is to viewData that's already passed in. If it's above a threshold it shows these solutions first, in order of most similar.
- After that, it offers to make the variable optional in the blade template by adding `?? ''`

Screen Shot 2019-09-01 at 14 34 23

## The Solutions

- Both solutions start by loading the blade template, and making an in-memory change
- The original and resulting blade templates are then compiled to PHP using the Blade Compiler
- The PHP is then tokenized
- We modify the tokenized original to our expected output
- The actual tokenized output is compared to the proposed resulting file, if they don't match, the solution action is never offered to the user, but the suggestion text is.

This is to ensure we never break a blade template, it also allows us to build upon the basic find and replace logic in the future.

## Automated Tests

There are some automated tests around this, we can expand to include more difficult cases.


freekmurze:
This looks awesome. I’ll review it soon. Thanks!

MrRio:
Thanks! :)

MrRio:
I'm happy with this if you are, will use draft PRs in future

src/SolutionProviders/UndefinedVariableSolutionProvider.php
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        return $this->getNameAndView($throwable) !== null;
+    }
+
+    public function getSolutions(Throwable $throwable): array
+    {
+        $solutions = [];
+
+        extract($this->getNameAndView($throwable));
```
mpociot: If this method returns `null`, the `$variableName` variable won't be available and the code won't work.
Maybe you should make sure that the `getNameAndView` method always returns those two variables that you need.

MrRio: It's called in `canSolve`, if it returns null it then `getSolutions` doesn't run. I can add an `isset`/return early to defend against that assumption changing in the future.

src/SolutionProviders/UndefinedVariableSolutionProvider.php
```diff
@@ -0,0 +1,78 @@
...
+        })->sortByDesc('match')->filter(function ($var, $key) {
+            return $var['match'] > 40;
+        })->keys()->map(function ($suggestion) use ($variableName, $viewFile) {
+            return new SuggestCorrectVariableNameSolution($variableName, $viewFile, $suggestion);
+        })->map(function ($solution) {
+            // If the solution isn't runnable, then just return the suggestions without the fix
+            if ($solution->isRunnable()) {
+                return $solution;
+            } else {
+                return BaseSolution::create($solution->getSolutionTitle())
```
freekmurze: Move this statement out of the `if`, so we can lose that `else`.

src/SolutionProviders/UndefinedVariableSolutionProvider.php
```diff
@@ -0,0 +1,78 @@
...
+                    ->setSolutionDescription($solution->getSolutionActionDescription());
+            }
+        })->toArray();
+
+        $optionalSolution = new MakeViewVariableOptionalSolution($variableName, $viewFile);
+        if ($optionalSolution->isRunnable()) {
```
freekmurze: Let's use a ternary here, so we can lose the `if` construct.

src/SolutionProviders/UndefinedVariableSolutionProvider.php
```diff
@@ -0,0 +1,78 @@
...
+            $solutions[] = $optionalSolution;
+        } else {
+            $solutions[] = BaseSolution::create($optionalSolution->getSolutionTitle())
+                ->setSolutionDescription($optionalSolution->getSolutionActionDescription());
+        }
+
+        return $solutions;
+    }
+
+    private function getNameAndView(Throwable $throwable)
```
freekmurze: Make this protected and add a return type hint.

src/Solutions/MakeViewVariableOptionalSolution.php
```diff
@@ -0,0 +1,95 @@
...
+        if ($output !== false) {
+            file_put_contents($parameters['viewFile'], $output);
+        }
+    }
+
+    public function makeOptional(array $parameters = [])
+    {
+        $originalContents = file_get_contents($parameters['viewFile']);
+        $newContents = str_replace('$'.$parameters['variableName'], '$'.$parameters['variableName']." ?? ''", $originalContents);
+        // Compile blade, tokenize
```
freekmurze: The code is clear enough, remove the comment

src/Solutions/MakeViewVariableOptionalSolution.php
```diff
@@ -0,0 +1,95 @@
...
+        $originalTokens = token_get_all(Blade::compileString($originalContents));
+        $newTokens = token_get_all(Blade::compileString($newContents));
+        // Generate what we expect the tokens to be after we change the blade file
+        $expectedTokens = [];
```
freekmurze: Move this piece of functionality to it's own function with a good name, so we don't need a comment anymore.

freekmurze: Do the same for the other solutions in this PR.

MrRio:
Thanks for the feedback. All sorted!

MrRio:
Good to merge?

src/Solutions/SuggestCorrectVariableNameSolution.php
```diff
@@ -0,0 +1,105 @@
...
+
+    public function run(array $parameters = [])
+    {
+        $output = $this->fixTypo($parameters);
+        if ($output !== false) {
+            file_put_contents($parameters['viewFile'], $output);
+        }
+    }
+
+    protected function fixTypo(array $parameters = [])
```
freekmurze: Add a return type hint here

src/SolutionProviders/UndefinedVariableSolutionProvider.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        $solutions = [];
+
+        extract($this->getNameAndView($throwable));
+
+        if (! isset($variableName)) {
+            return [];
+        }
+
+        $solutions = collect($throwable->getViewData())->map(function ($value, $key) use ($variableName) {
```
freekmurze: Put this is a `determineCorrectVariableSolutions()` function (or something with a better name.

src/SolutionProviders/UndefinedVariableSolutionProvider.php
```diff
@@ -0,0 +1,72 @@
...
+        })->keys()->map(function ($suggestion) use ($variableName, $viewFile) {
+            return new SuggestCorrectVariableNameSolution($variableName, $viewFile, $suggestion);
+        })->map(function ($solution) {
+            return $solution->isRunnable()
+                ? $solution
+                : BaseSolution::create($solution->getSolutionTitle())
+                    ->setSolutionDescription($solution->getSolutionActionDescription());
+        })->toArray();
+
+        $optionalSolution = new MakeViewVariableOptionalSolution($variableName, $viewFile);
```
freekmurze: Put this in a function named `determineMakeVariableOptional` (or something with an even better name).

src/Solutions/SuggestCorrectVariableNameSolution.php
```diff
@@ -0,0 +1,105 @@
+<?php
+
+namespace Facade\Ignition\Solutions;
+
+use Illuminate\Support\Facades\Blade;
+use Facade\IgnitionContracts\RunnableSolution;
+
+class SuggestCorrectVariableNameSolution implements RunnableSolution
+{
+    private $variableName;
```
freekmurze: Add type docblocks here.

src/Solutions/SuggestCorrectVariableNameSolution.php
```diff
@@ -0,0 +1,105 @@
...
+    public function getDocumentationLinks(): array
+    {
+        return [];
+    }
+
+    public function getSolutionActionDescription(): string
+    {
+        $path = str_replace(base_path().'/', '', $this->viewFile);
+        $output = [
+            'Did you mean `$'.$this->suggested.'`?',
```
freekmurze: Use string interpolation here.

src/Solutions/SuggestCorrectVariableNameSolution.php
```diff
@@ -0,0 +1,105 @@
...
+    public function getRunParameters(): array
+    {
+        return [
+            'variableName' => $this->variableName,
+            'viewFile' => $this->viewFile,
+            'suggested' => $this->suggested,
+        ];
+    }
+
+    public function isRunnable(array $parameters = [])
```
freekmurze: Add return type hint here.

src/Solutions/SuggestCorrectVariableNameSolution.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        return $this->fixTypo($this->getRunParameters()) !== false;
+    }
+
+    public function run(array $parameters = [])
+    {
+        $output = $this->fixTypo($parameters);
+        if ($output !== false) {
```
freekmurze: I often want the successful situation at the bottom of a function and bailing out situations above it.

So change this to 

```
if ($output === false) {
   return;
}

file_put_contents($parameters['viewFile'], $output);
```

src/Solutions/SuggestCorrectVariableNameSolution.php
```diff
@@ -0,0 +1,105 @@
...
+            'suggested' => $this->suggested,
+        ];
+    }
+
+    public function isRunnable(array $parameters = [])
+    {
+        return $this->fixTypo($this->getRunParameters()) !== false;
+    }
+
+    public function run(array $parameters = [])
```
freekmurze: Add `void` typehint.

src/Solutions/SuggestCorrectVariableNameSolution.php
```diff
@@ -0,0 +1,105 @@
...
+        $output = $this->fixTypo($parameters);
+        if ($output !== false) {
+            file_put_contents($parameters['viewFile'], $output);
+        }
+    }
+
+    protected function fixTypo(array $parameters = [])
+    {
+        // Make sure suggested variable is valid alpha-numeric with underscore, or return false
+        if (! preg_match('/^[a-zA-Z]+[a-zA-Z0-9_]+$/', $parameters['suggested'])) {
```
freekmurze: Put this test in a function name `isAlphaNumericWithUnderscore` so we can lose the comment.

src/Solutions/SuggestCorrectVariableNameSolution.php
```diff
@@ -0,0 +1,105 @@
...
+
+        return $newContents;
+    }
+
+    protected function generateExpectedTokens(array $originalTokens, string $variableName, string $suggested): array
+    {
+        $expectedTokens = $originalTokens;
+        foreach ($expectedTokens as $key => $token) {
+            if ($token[0] === T_VARIABLE && $token[1] === '$'.$variableName) {
+                $expectedTokens[$key][1] = '$'.$suggested;
```
freekmurze: Use string interpolation here.

src/Solutions/SuggestCorrectVariableNameSolution.php
MrRio: Can't here because it's either a string or false

src/SolutionProviders/UndefinedVariableSolutionProvider.php
MrRio: Is now in findOptionalVariableSolution

MrRio:
This is good to merge. 👍

mpociot:
Great job on this PR! Thank you!

PR #39, opened 2019-09-01
Update gitignore and styleci

DonJorgen:
- Is easy to follow follow and more logical to use alphabetical ordering for imports

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 composer.lock
 .php_cs.cache
 .phpunit.result.cache
+/.idea/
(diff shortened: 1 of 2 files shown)


freekmurze:
I'm going to pass on this one. The current StyleCI config is what both Spatie and Beyond Code are using for their packages. The `.idea` belongs in a global `.gitignore`.

DonJorgen:
So I will make a new one to fix phpcs, I thought the idea was to have like this because of this rule :) 

```
        'ordered_imports' => ['sortAlgorithm' => 'alpha'],
```

PR #41, opened 2019-09-01
Prevent filter and duplicated code

DonJorgen:


src/Actions/ShareReportAction.php
```diff
@@ -86,11 +95,9 @@ protected function filterContextItems(array $contextItems): array
 
     protected function removeRequestInformation(array $contextItems): array
     {
-        Arr::forget($contextItems, 'request');
-        Arr::forget($contextItems, 'request_data');
-        Arr::forget($contextItems, 'headers');
-        Arr::forget($contextItems, 'session');
-        Arr::forget($contextItems, 'cookies');
+        foreach ($this->forgeables as $forgeable) {
```
freekmurze: I like most of the changes except this one. There's seemingly no reason why the array is an instance variable. Let's inline it.

freekmurze:
I've decided not to merge this in, but I'd manually use some of the suggestions in your PR. Thanks!

DonJorgen:
Will try to grab more free time tovdig a bit more on the package ... Great talk BTW :)

PR #42, opened 2019-09-01
Automatically fix incorrect view names by either creating the file, or changing it in the controller

MrRio:
# This PR adds a solution for fixing missing views

Screen Shot 2019-09-01 at 17 47 03

## The Solution Provider

- It uses the existing logic to see if there is a close enough view name to suggest. If there is, it will offer to change it for you.
- Otherwise, it will offer to create the view file for you

Screen Shot 2019-09-01 at 17 47 28

## Update View Name Solution

- It looks through the stack trace for the first file within the users' app/ directory, this will generally be the controller
- It then loads the controller into memory, and converts it into an array of PHP tokens
- We change string names that match the missing view, and change this in the array of tokens
- We retokenize the output, and compare what we expected with what we got, if this doesn't work, the user isn't offered an automatic fix

## Create View File Solution

- It explodes the view by `.`
- recursively makes folders in the first scanned view path for the app
- Touches the full file path to create an empty file

## Automated Tests

There are some automated tests around this, we can expand to include more difficult cases.

## TODO

- [x] Make sure we only update the string if it's inside a call to view()
- [x] Test for create file solution


freekmurze:
Seems very promising, keep up the good work!

For future PRs you want to work on, create a draft PR: https://github.blog/2019-02-14-introducing-draft-pull-requests/

MrRio:
@freekmurze Hey, this is good to merge :-)

MrRio:
CC @freekmurze 👀

MrRio:
Hey @mpociot — hoping to get these merged soon, let me know if there's anything I can do to make it easier, I know you're both very busy!

MrRio:
Also updated this branch with the latest master, tests passing and StyleCI is happy.

MrRio:
CC @freekmurze @mpociot

freekmurze:
Some people have raised concerns about ignition changing source code, so we're going to hold off this PR for now.

Thanks for your work on it though.

PR #43, opened 2019-09-01
Change assert null to assert false

DonJorgen:
Change from assert null to assert false, to follow the same pattern as the rest of the asserts


freekmurze:
Thanks!

PR #45, opened 2019-09-01
Change import rules to match styleci

DonJorgen:
- Ref: https://github.com/facade/ignition/pull/39


freekmurze:
Meanwhile, I've just removed that file. We'll rely on StyleCI.

Issue #27, opened 2019-08-31
Prefix for file paths (for Homestead users)


klaasgeldof:
If you use Homestead, the real paths of your code files look something like this:

`/home/vagrant/code/myproject/app/Http/Controllers/MyController.php`

But on my computer, that isn't the real path to the file. Instead, this is the correct path:

`/Users/klaas/Documents/Code/myproject/app/Http/Controllers/MyController.php`

So if I click the pencil icon on the Ignition page, the file is not found.

I suggest you can add a prefix in the Ignition config file, like this:

```php
[
    'path_prefix' => '/Users/klaas/Documents/Code/myproject',
]
```

If defined, it gets prefixed to the paths, which can be done like this for example:

```php
if ($prefix = config('ignition.path_prefix')) {
    $path = $prefix.substr($path, strlen(base_path()));
}
```

simon511000:
I have exactly the same problem as @klaasgeldof, his fix is perfect!

freekmurze:
If you have the time, feel free to send a PR for this.

Elgoist:
Working on a PR. Docker Development is broken as well.

freekmurze:
`remote-sites-path` and `local-sites-path` config keys have been value to the config file that'll help you solve this. Thanks for the PR @mikebronner 👍

mikebronner:
🤙

klaasgeldof:
Great, thanks! 😃

Purist comment: config keys should be snake case, so:

`remote-sites-path` => `remote_sites_path`
`local-sites-path` => `local_sites_path`

Now there are mixed cases in the config file.

freekmurze:
@mpociot is on it!

PR #47, opened 2019-09-01
Add VSCODE-INSIDERS edit link; fix Homestead edit link.

mikebronner:
This PR solves two problems:
- enable editing URLs of projects running on Homestead #27 
- enable editing URLs when using VSCODE-INSIDERS.

In hindsight I probably should have split them in two PRs, but they were so closely intertwined that I had to fix one to do the other.


config/ignition.php
```diff
@@ -9,11 +9,26 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | If you are using homestead, it will be necessary to specify your path
+    | mapping, just like in your Homestead.yaml file. Leaving either one, or
+    | both of these empty or null will not trigger the homestead URL changes.
+    |
+    | 'homestead-sites-path' is the base path inside homestead, f.e. `/home/vagrant/Code`
+    | 'local-sites-path' is the base path on your host computer, f.e. `/Users/<name>/Code` or `C:\Users\<name>\Documents\Code`
+    */
+    'homestead-sites-path' => env('IGNITION_HOMESTEAD_SITES_PATH', ''),
```
johanvanhelden: I use a custom docker solution, maybe it's better to rename this to `server-path` and `local-path`, so it's not colored as to what server-setup you are using.

simon511000:
I just tested and it works perfectly, good job @mikebronner !

config/ignition.php
mikebronner: Good -pint, thanks for the feedback. I'll get that changed in a bit.

johanvanhelden: Awesome, thank you so much! And maybe also make the description more generic. So instead of making everything homestead specific you could call it "Server path mapping". And "If you use a server setup with for example Vagrant or Docker, it will be necessary to..."

mikebronner:
Done, should be ready to go now. :) Thanks for the feedback, guys.

johanvanhelden:
It looks perfect to me now. Thank you so much for your time and making this PR! 

Hopefully it gets merged and tagged soon so we can use the editor functionality!

PR #48, opened 2019-09-01
[WIP] Suggest classes based on text similarity, with a weight towards Facades

MrRio:
Draft PR to support matching similar (preferably exact) class names to import, with a weighting towards Facades.

Screen Shot 2019-09-02 at 00 01 41

## Todo

- [ ] Add fixer to add use statement
- [ ] Use parser tokens to correctly place `use` statements 
- [ ] If coding style appears to be ordering the 'use' statements in length order, then continue to match that style
- [ ] Change weighting of classes imported to: Users' application first, Facades, the rest of Laravel, 3rd party deps 
- [ ] Automated tests


Gummibeer:
Just as an idea: it's possible to have a single card show multiple solutions. Wouldn't it be an idea to show "all" matching classes?

freekmurze:
Closing due to inactivity.

Issue #44, opened 2019-09-01
Introduce a way to disable certain solutions


freekmurze:
If somebody is not happy with a certain solution provider there should be a way to turn it off. 

Let's introduce a config key that can hold a list of ignores solutions.

percymamedy:
Hello @freekmurze 

Did an attempt to ignore certain solutions, check PR https://github.com/facade/ignition/pull/49

Cheers !!

freekmurze:
We’ll continue this conversation in #49

PR #49, opened 2019-09-02
Issue #44: Disable certain solutions

percymamedy:
- Added a config to disable solutions providers.


src/SolutionProviders/SolutionProviderRepository.php
```diff
@@ -47,7 +47,8 @@ public function getSolutionsForThrowable(Throwable $throwable): array
 
         $providerSolutions = $this->solutionProviders
             ->filter(function (string $solutionClass) {
-                return in_array(HasSolutionsForThrowable::class, class_implements($solutionClass));
+                return in_array(HasSolutionsForThrowable::class, class_implements($solutionClass)) &&
+                       ! in_array($solutionClass, config('ignition.disabled_solutions'));
```
riasvdv: Might want to change `config('ignition.disabled_solutions')` to `config('ignition.disabled_solutions', [])` for when the config isn't set.

freekmurze:
Thanks!

src/SolutionProviders/SolutionProviderRepository.php
freekmurze: I'm thinking it will always be set since we merge the published config with the unpublished one.

PR #52, opened 2019-09-02
Add ability to disable share button

mpociot:


config/ignition.php
```diff
@@ -24,4 +24,19 @@
...
+
+
+
+    /*
+    |--------------------------------------------------------------------------
+    | Sharing
+    |--------------------------------------------------------------------------
+    |
+    | Ignition allows you to share your local errors with colleagues or people around the world.
+    | Sharing errors is completely free and does not require an account on Flare.
```
freekmurze: Maybe also mention that it doesn't share errors automatically, it's always a manual action.

Issue #24, opened 2019-08-31
Having no context ends in an error


stefanbauer:
If there is no context, it ends in a JS error. I am not sure if it only appears if there is no context or what that error is saying about that URL. Anyway.. that said, the view ends in a "gray space". Maybe it makes sense to use their also white background with a custom message that there is no context given or sth. like that.

For reproducing:

- Latest 5.8 version
- Install Ignition
- Use an exception as closure in the welcome route
- Head over to the "Context" tab

![image](https://user-images.githubusercontent.com/3192662/64062364-b409c800-cbe5-11e9-97a0-b9e6d719679a.png)

axit-joost:
I had this too, until I pushed to a remote git repo. Before that, Context tab remained blank and threw that error on the console.

AdrianMrn:
This happens because we were trying to generate a url of the git repo even if the project isn't using git. This PR fixes it: https://github.com/facade/ignition/pull/57

freekmurze:
This is fixed in the latest version.

PR #57, opened 2019-09-02
Fix context tab crash when not using git (#24)

AdrianMrn:
We were trying to get a repo url even when a project wasn't using git.


freekmurze:
Thanks!

PR #58, opened 2019-09-02
Show nice error message for merge conflicts

MrRio:
Screen Shot 2019-09-02 at 17 28 04

This is a relatively simple Solution Provider that shows a pretty error message for a merge conflict.

I don't think we offer an automatic solution here as the results are potentially quite drastic for a single accidental click. 

## Notes:

* ~There's a workaround in place as it's impossible to turn off linting in style-ci for any .php file, but we're going to want to test with a real actually broken file. We copy this into place for the test.~ 

## Todo:
- [x] Automated test


src/SolutionProviders/MergeConflictSolutionProvider.php
```diff
@@ -0,0 +1,45 @@
...
+{
+    public function canSolve(Throwable $throwable): bool
+    {
+        if (! ($throwable instanceof FatalThrowableError || $throwable instanceof \ParseError)) {
+            return false;
+        }
+
+        if (Str::startsWith($throwable->getMessage(), 'syntax error, unexpected \'<<\'')) {
+            $file = file_get_contents($throwable->getFile());
+            if (strpos($file, '<<<<<<<') !== false &&
```
freekmurze: Could you break this up into multiple `if `statements / early returns?

MrRio: Sure!

freekmurze:
This is awesome! 👍

src/SolutionProviders/MergeConflictSolutionProvider.php
```diff
@@ -0,0 +1,49 @@
...
+        return true;
+    }
+
+    public function getSolutions(Throwable $throwable): array
+    {
+        $file = file_get_contents($throwable->getFile());
+        preg_match('/\>\>\>\>\>\>\> (.*?)\n/', $file, $matches);
+        $source = $matches[1];
+        $folder = basename($throwable->getFile());
+        $target = trim(`cd $folder; git branch | grep \* | cut -d ' ' -f2`);
```
mpociot: What happens if the command inside the `trim` call fails? For example because this happens on a windows machine? Should this be wrapped in a try-catch block?
The solution should also change if the target could not be automatically detected.

MrRio: I can't make the call throw an exception, but it can return a blank string.

viezel:
very nice 👏

src/SolutionProviders/MergeConflictSolutionProvider.php
```diff
@@ -0,0 +1,60 @@
...
+use Illuminate\Support\Str;
+use Facade\IgnitionContracts\BaseSolution;
+use Facade\IgnitionContracts\HasSolutionsForThrowable;
+use Symfony\Component\Debug\Exception\FatalThrowableError;
+
+class MergeConflictSolutionProvider implements HasSolutionsForThrowable
+{
+    public function canSolve(Throwable $throwable): bool
+    {
+        if (! ($throwable instanceof FatalThrowableError || $throwable instanceof \ParseError)) {
```
freekmurze: Nitpick: could you import `ParseError` ?

src/SolutionProviders/MergeConflictSolutionProvider.php
```diff
@@ -0,0 +1,60 @@
...
+
+        return [
+            BaseSolution::create("Merge conflict from branch '$source' into $target")
+                ->setSolutionDescription('You have a Git merge conflict. To undo your merge do `git reset --hard HEAD`'),
+        ];
+    }
+
+    private function getCurrentBranch($folder)
+    {
+        $branch = "'".trim(`cd $folder; git branch | grep \* | cut -d ' ' -f2`)."'";
```
freekmurze: What happens if this is executed in a project that doesn't use `git`. Or when `git` is not available?

tests/Solutions/MergeConflictSolutionProviderTest.php
```diff
@@ -0,0 +1,31 @@
...
+
+        View::addLocation(__DIR__.'/../stubs/views');
+    }
+
+    /** @test */
+    public function it_can_solve_merge_conflict_exception()
+    {
+        copy(__DIR__.'/../stubs/Controllers/GitConflictController.stub', __DIR__.'/../stubs/Controllers/GitConflictController.php');
+        try {
+            app(\Facade\Ignition\Tests\stubs\Controllers\GitConflictController::class);
```
freekmurze: Import this class.

src/SolutionProviders/MergeConflictSolutionProvider.php
MrRio: Sure

src/SolutionProviders/MergeConflictSolutionProvider.php
MrRio: Any error inside `` returns a blank string, which then makes the message say "from branch 'xyz' into current branch"

MrRio:
@freekmurze All good?

freekmurze:
Thank you for this! 👍

MrRio:
No problem!

PR #64, opened 2019-09-03
Disable solution execution

mpociot:
This PR adds the ability to explicitly disable the support for runnable solutions. By default we allow runnable solutions while the application is in debug mode, but in some cases you might want to disable this even then.


freekmurze:
Very nice!

PR #69, opened 2019-09-03
support monologger 2

freekmurze:


freekmurze:
This isn't needed as `void` was introduced in PHP 7.1

PR #77, opened 2019-09-03
On first run after upgrading an application to Laravel 6, error

CedNet:
It will throw an exception on this line.

"The bug": If you don't run vendor publication of the default configs, this will throw an error rendering the page with the default symfony debugger.

No need to merge this exactly, but it solves the problem :)


freekmurze:
I'm going to already merge it in and tag a new release, so you and others don't have this problems anymore, and we'll polish this solution tomorrow.

Thank you!

PR #83, opened 2019-09-04
Suggest running migrations when a column is missing

MrRio:
This is a slight tweak of the existing 'missing table' solution provider. Generally, when you're missing a column, you're also behind on your migrations.


src/SolutionProviders/MissingColumnSolutionProvider.php
```diff
@@ -0,0 +1,37 @@
...
+use Illuminate\Database\QueryException;
+use Facade\Ignition\Solutions\RunMigrationsSolution;
+use Facade\IgnitionContracts\HasSolutionsForThrowable;
+
+class MissingColumnSolutionProvider implements HasSolutionsForThrowable
+{
+    /**
+     * See https://dev.mysql.com/doc/refman/8.0/en/server-error-reference.html#error_er_bad_field_error.
+     */
+    const MYSQL_BAD_TABLE_CODE = '42S22';
```
mpociot: Can you rename the constant to `BAD_FIELD_CODE`?

src/Solutions/RunMigrationsSolution.php
```diff
@@ -7,8 +7,19 @@
...
+        $this->customTitle = $customTitle;
+    }
+
     public function getSolutionTitle(): string
     {
+        if (isset($this->customTitle)) {
+            return $this->customTitle;
+        }
+
         return 'A table was not found';
```
mpociot: Can you change this to always return the title from the constructor? Then the existing table not found solution can also just provide this text and both solution providers behave the same way and we get rid of this if statement.

MrRio: Cool. I will need to fallback the argument to an empty string for when the class is reconstructed outside the solution provider to call `run()`

MrRio:
Got a very exciting idea for a progression on this when merged 👀

freekmurze:
Thank you!

PR #84, opened 2019-09-04
Lookup the first exception entry when linking to Telescope

mpociot:
Right now we simply use the first entry that is in the Telescope entry queue. This might not be the correct one though, as other entries could be there - like logs or queues.

This PR fixes this by checking that the entry is for an exception.

### DIFF
--- a/src/ErrorPage/ErrorPageViewModel.php
+++ b/src/ErrorPage/ErrorPageViewModel.php
@@ -10,6 +10,7 @@ use Facade\FlareClient\Report;
 use Laravel\Telescope\Telescope;
 use Facade\Ignition\IgnitionConfig;
 use Illuminate\Contracts\Support\Arrayable;
+use Laravel\Telescope\IncomingExceptionEntry;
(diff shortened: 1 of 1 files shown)


src/ErrorPage/ErrorPageViewModel.php
```diff
@@ -61,7 +62,15 @@ public function telescopeUrl(): ?string
                 return '';
             }
 
-            $telescopeEntryId = (string) Telescope::$entriesQueue[0]->uuid;
+            $telescopeEntry = collect(Telescope::$entriesQueue)->first(function ($entry) {
+                return $entry instanceof IncomingExceptionEntry;
+            });
+
+            if (is_null($telescopeEntry)) {
+                return '';
```
freekmurze: The return type hint is nullable here. That makes me thinking that the calling code could expect `null` instead of empty string.

freekmurze:
Seems ok, just one remark to look at. 👍

PR #89, opened 2019-09-04
Remove quotation from git commitUrl

felipepodesta:
Remove quotation from git commitUrl on Git Context Tab.


mpociot:
Thank you for the PR. Can you show me an example of where this was wrong?

felipepodesta:
> Thank you for the PR. Can you show me an example of where this was wrong?

Yes, the url is coming with single quotes.

image

PR #91, opened 2019-09-04
Add SolutionProvider for missing controllers

jerodev:
Add a solution provider for missing controllers.


MrRio:
Typo in description “You route is pointing” should be Your

mpociot:
How can I test this solution provider?

Doing a `Route::get('unknown-controller', 'SomeController@method');` does not return that solution and creates a `Class App\Http\Controllers\SomeController does not exist` ReflectionException

jerodev:
I create a route that goes to a controller that does not exist and visit that route in the browser.
May I ask how you tested the solution provider? Maybe I forgot another case where this might happen?

image

mpociot:
@jerodev It probably depends on how you define the route.

This is my `web.php` file:

```php
Route::get('unknown-controller', 'SomeController@method');
```

And then I visit `/unknown-controller`.

jerodev:
It seems the `Exception` for missing controllers has been changed in Laravel 6.0, which I was testing with.

https://github.com/laravel/framework/commit/f336179c715d81c9a8e787576935278d3e2c499d#diff-52441e04e14c52275cd5d09e8f958981

Laravel 6.0 and onward throws a `BindingResolutionException`, whereas previous versions throw a `ReflectionException`.

I will update the code to support both exception types.

jerodev:
I have added support for `ReflectionException` for Laravel 5.x. 
This should solve the problem where the solution did not show @mpociot.

src/Solutions/CreateControllerSolution.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        $controller = $this->getControllerPath();
+
+        return "Your route is pointing to a controller that does not exist. You can create the controller using `php artisan make:controller $controller`.";
+    }
+
+    public function getDocumentationLinks(): array
+    {
+        return [
+            'Basics: Controller docs' => 'https://laravel.com/docs/6.0/controllers',
```
freekmurze: Use the first character of `\Illuminate\Foundation\Application::VERSION` instead of hardcoding `6`.

freekmurze:
Closing du to inactivity.

Issue #87, opened 2019-09-04
Ignition not working on Laravel 6 [ShareReportController not defined]


gsabater:
Hello,

Looks like the ignition page is not working on an updated project.
The project was 5.8 and now is 6.0 (so it's not related to #20)
Ignition is updated to 1.4.8

The framework displays this page, and no issues are registered on the logs
![ignition](https://user-images.githubusercontent.com/4948172/64251786-999f5980-cf19-11e9-9930-d2b56b896f30.png)

mpociot:
Could it be that your routes are cached? The ignition package introduces some new routes that you don't have loaded.

Please try clearing your route cache using `php artisan route:clear`.

gsabater:
Hello @mpociot 

I have tested your suggestion, and executed `php artisan route:clear`, `php artisan cache:clear`, `php artisan config:clear` and `php artisan view:clear` testing after each clear without success.

Also, I have made a `composer dump-autoload -o` and finally I have deleted manually `bootstrap>cache>packages.php and services.php`

Everything yields the same result.
Salut, thanks

gsabater:
I'd like to add that to check the error page, I have manually removed `use Illuminate\Support\Str;` in a middleware loaded from boot in a serviceprovider. The service provider is loaded in `config.php > app > providers`.

mpociot:
I'm trying to create a reproducable version of this error, but couldn't achieve it yet.

Where in the providers array is the service provider loaded? How do you push the middleware?

gsabater:
Provider:
```
'providers' => [

        //
        // Laravel Framework Service Providers...
        // -------------------------------------
        ...
        Illuminate\Translation\TranslationServiceProvider::class,
        Illuminate\Validation\ValidationServiceProvider::class,
        Illuminate\View\ViewServiceProvider::class,

        //
        // Bonda service providers...
        // -------------------------------------
        Bonda\Core\CoreServiceProvider::class,
        ...
```

And the middleware is loaded like this in the CoreServiceProvider.php
```
public function boot()
    {
        app('Illuminate\Contracts\Http\Kernel')
            ->pushMiddleware(\Bonda\Core\Middleware\EnvironmentWarning::class);
    }
```

[...]

PR #93, opened 2019-09-04
Use a middleware to protect ignition routes

mpociot:
By always registering our routes we avoid issues like #87 
The middleware then takes care of choosing whether or not the route should be reachable.

### DIFF
--- a/src/Http/Middleware/IgnitionEnabled.php
+++ b/src/Http/Middleware/IgnitionEnabled.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Facade\Ignition\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+
+class IgnitionEnabled
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param  Request  $request
+     * @param  Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        if (! $this->ignitionEnabled()) {
+            abort(404);
+        }
+
+        return $next($request);
+
+    }
+
+    protected function ignitionEnabled(): bool
+    {
+        return config('app.debug');
+    }
+}
--- a/src/IgnitionServiceProvider.php
+++ b/src/IgnitionServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace Facade\Ignition;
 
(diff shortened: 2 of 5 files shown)


src/Http/Middleware/IgnitionEnabled.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Facade\Ignition\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+
+class IgnitionEnabled
```
freekmurze: Maybe add a test for this middleware, so we're 100% sure it works correctly.

src/IgnitionServiceProvider.php
```diff
@@ -117,19 +118,17 @@ protected function registerViewEngines()
...
+            'prefix' => config('ignition.housekeeping_endpoint_prefix', '_ignition'),
+            'middleware' => [IgnitionEnabled::class],
+        ], function () {
+            Route::get('health-check', HealthCheckController::class);
+            Route::post('execute-solution', ExecuteSolutionController::class);
+            Route::post('share-report', ShareReportController::class);
+
+            Route::get('scripts/{script}', ScriptController::class);
+            Route::get('styles/{style}', StyleController::class);
+        });
```
franzliedke: Shouldn't routes be registered in `boot()`? That goes for almost every service in the IoC container - don't resolve them until all other service providers had a chance to modify / extend / swap them.

mpociot: Yeah, in our case this would introduce some issues if an exception would occur in a service providers boot method prior to Ignition.
Because our whoops handler would already be registered, but the routes would not. That's why we are registering the routes here too.

franzliedke: What about `$this->app->resolving('router', function () { ... });`?

freekmurze: Marcel and I discussed this. We don't think it is necessary. The router will always have been registered when our provider gets called.

franzliedke: The point is not so much your provider (although you shouldn't rely on the provider ordering). You also should be a good citizen and allow other packages to swap out the router / extend it in some way, *before* you access it.

That's at least 50% of the point of the IoC container and contracts, as I understand them. :wink:

PR #94, opened 2019-09-04
Automagically fix missing semicolons, uses tokenization and only suggests with correct linting

MrRio:
# A Solution for Fixing Missing Semicolons

Screen Shot 2019-09-04 at 22 18 53

# How it works

1. First it takes the unexpected token from the error message
2. Then it tokenizes the file in which the syntax error occurred
3. It reverses the order of the tokenization, and loops over to find the unexpected token
4. If then waits until the next whitespace token, and inserts that, THEN a semicolon
5. The file is regenerated in memory with the proposed fix
6. If this in-memory representation of the fix now passes PHP linting, it is offered as a saveable solution

Todo:

- [x] Refactor semicolon insertion slightly
- [x] Add automated tests


MrRio:
@freekmurze This is all good to review 👍

MrRio:
Have updated with the latest master, tests passing and StyleCI is happy.

MrRio:
CC @freekmurze @mpociot

freekmurze:
Thanks for the PR, we'll review it soon.

freekmurze:
Could you add a test that proves that the `;` is added correctly?

freekmurze:
Closing due to inactivity. Feel free to open a new PR should you want to continue working on this.

PR #95, opened 2019-09-05
Use HTTPS link for Gravatar

mbardelmeijer:
Prevent mixed content warnings

`Mixed Content: The page at 'https://flareapp.io/occurrence/XXXXXX' was loaded over HTTPS, but requested an insecure image 'http://www.gravatar.com/avatar/XXXXXX.jpg?s=80'. This content should also be served over HTTPS.`


mpociot:
thank you!

PR #102, opened 2019-09-05
build all Laravel versions on php 7.4snapshot and allow failures

Gummibeer:
This PR adds all Laravel versions in lowest and stable config and PHP 7.4 snapshot to the build matrix and allows failures.


freekmurze:
Thanks!

olivernybroe:
That was a fast pr after my tweet :O

Gummibeer:
@olivernybroe I beliefe that I worked on it before your tweet. Or at least before I've seen it.^^
Have added the same to https://github.com/spatie/laravel-enum/pull/20 - so I thought it would be nice to also add it here because of Freeks tweet that they test with everything. 😄

PR #109, opened 2019-09-05
add composer support section

Gummibeer:
and normalize `composer.json` using https://github.com/localheinz/composer-normalize


mpociot:
Thanks!

PR #110, opened 2019-09-05
Fix registering multiple solution providers

jasonvarga:
When using `registerSolutionProviders`, nothing was happening because the property wasn't being set.


mpociot:
thank you!

freekmurze:
👍

Issue #104, opened 2019-09-05
Ignition doesn't open file in phpstorm


emargareten:
I set up my environment according to https://flareapp.io/docs/ignition-for-laravel/configuration but it does not open my editor to the file/line

antonkomarev:
Duplicate #33

PR #112, opened 2019-09-06
Cross-platform PHPStorm editor URL

### LINKED ISSUE #33: PHPStorm editor links only work on Mac
The `phpstorm://` scheme currently only works on Mac, it requires a third party plugin to work on Windows/Linux, like [aik099/PhpStormProtocol](https://github.com/aik099/PhpStormProtocol). As mentioned in #27 this also doesn't work when Homestead is used.

IntelliJ also exposes an REST API at `http://localhost:63342/`, see https://www.develar.org/idea-rest-api/#api-Platform-about but then Ignition has to make a get request to `http://localhost:63342/api/file?file=routes/web.php&line=17` for example. 
<details><summary>Remove the warning popups</summary>
<p>

Settings -> Build, Execution, Deployment -> Debugger -> Check "Allow unsigned requests" under Built-in server removes the warning popups

image

</p>
</details>

Another option is to use the `jetbrains://` scheme, that's included in Jetbrains Toolbox. The scheme works on all platforms, however I couldn't get that scheme to open any files, only errors.

So since PHPStorm currently doesn't support opening files that well, would it be worth investing time in a PR which makes for a better cross-platform compatibility? Or maybe just make it possible to disable the "open file" feature and adding notes to the config file or docs?

antonkomarev:
Resolves #33 #104 

This solution was tested in Laravel Whoops Editor package:
https://github.com/cybercog/laravel-whoops-editor/blob/master/config/whoops-editor.php#L48

PHPStorm fix:
```
http://localhost:63342/api/file/?file=%path&line=%line
```

### DIFF
--- a/resources/js/components/Shared/editorUrl.js
+++ b/resources/js/components/Shared/editorUrl.js
@@ -5,7 +5,7 @@ export default function editorUrl(config, file, lineNumber) {
         textmate: 'txmt://open?url=file://%path&line=%line',
         emacs: 'emacs://open?url=file://%path&line=%line',
         macvim: 'mvim://open/?url=file://%path&line=%line',
-        phpstorm: 'phpstorm://open?file=%path&line=%line',
+        phpstorm: 'http://localhost:63342/api/file/?file=%path&line=%line',
         idea: 'idea://open?file=%path&line=%line',
         vscode: 'vscode://file/%path:%line',
         'vscode-insiders': 'vscode-insiders://file/%path:%line',


antonkomarev:
It looks like there is a need to make asynchronous Ajax call. Otherwise blank page is opening on background.

mpociot:
@antonkomarev I think we will need to do ajax requests then too and our configuration should also contain the url and ajax key.
Can you update this PR for this?

antonkomarev:
Can start doing it not earlier than on Sunday.

kylekatarnls:
This fix worked for me on Windows.

Francismori7:
This opens DataGrip for me (also JetBrains)

antonkomarev:
@Francismori7 good to know that, thanks!

mbardelmeijer:
Whats the status of this PR? Would love to see this merged, if i can help in any way happy to help :)

antonkomarev:
As @Francismori7 noted - there is an issue with this implementation. This URL is used by all IntelliJ products and if you have more than 1 installed - there is no way to choose which one should be opened.

zschuessler:
@antonkomarev The original issue from @Francismori7 is likely because the local server was not configured in any of the IntelliJ products, and it's guessing what to use as a last resort.

Although requiring a user to configure that would allow PHPStorm to open, if you have multiple projects open the implementation may not know which PhpStorm project should get precedence. 

For opening the right project, a _Project Name_ config setting would allow PhpStorm to decide which to open:

https://stackoverflow.com/a/16290907/1060207

So here's an overview of what would need to change:

1. Allow user to set a Project Name variable for Ignition. (Maybe we could even guess it as a last resort?)
2. Set this project name in the API call to the local IntelliJ server (eg localhost:port/project-name/file)
3. Rely on user to have a properly configured PhpStorm setup, give documentation on how to do it.

zschuessler:
For anyone else looking for a quick fix until this gets attention, I have a quick hack for you.

**Step 1: Append Javascript to Page if It's an Exception**
in app/Exceptions/Handler.php:

```
public function render($request, Exception $exception)
    {
        if ('local' === app()->environment()) {
            echo view('debug.ignition-exception-page');
        }

        return parent::render($request, $exception);
    }
```

Obviously you'll want to create the debug/ignition-exception-page view, or handle the next bit however you want.

**Step 2: Replace all Protocol URLs to use PHPStorm API**
Next with our appended JavaScript, we switch all links starting with the phpstorm protocol to use the API instead.

```
<?php
/**
 * Fixes phpstorm:// protocol links Ignition uses, which don't work on Linux or Windows
 */
$projectPath = base_path();
?>


<script
    src="https://code.jquery.com/jquery-3.4.1.min.js"
    integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo="
    crossorigin="anonymous"></script>
<script>

    $(document).ready(function() {
        // Get all phpstorm:// protocol links
        var $links = $('a[href^=phpstorm]');

        $links.each(function(index) {
            var $link = $(this);
            var href = $link.attr('href');

            // Drop the protocol, which doesnt work on linux or windows
            href = href.replace('phpstorm://open?', 'http://localhost:63342/api/file?');

            // Add the project path so PHPstorm knows which window to make active
            href = href + '&project={{ $projectPath }}';
            $link.attr('href', href);

            // Send as an XHR request, so we don't redirect the user or have to open a new window
            $link.on('click', function(e) {
                e.preventDefault();

                $.get(href);
            })
        })
    });
</script>
```

You can use your own version of jQuery here (it's the cdn) and alter the project path as needed, but more than likely it'll always be your base_path() value.

Enjoy.

reppair:
Hi, any update on this one?

reppair:
An update on @zschuessler work around while waiting for a proper fix - the above code will do the trick but only for the last error in the stack trace. Clicking on a previous error in the stack will re-render the links with the original `phpstorm://` href. Here is an updated version that handles it via `MutationObserver`.

```
<?php
/**
 * Fixes phpstorm:// protocol links Ignition uses, which don't work on Linux or Windows
 */
$projectPath = base_path();
?>

<script
    src="https://code.jquery.com/jquery-3.4.1.min.js"
    integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo="
    crossorigin="anonymous"></script>
<script>

    function substituteLinks() {
        // Get all phpstorm:// protocol links
        var $links = $('a[href^=phpstorm]');

        $links.each(function(index) {
            var $link = $(this);
            var href = $link.attr('href');

            // Drop the protocol, which doesnt work on linux or windows
            href = href.replace('phpstorm://open?', 'http://localhost:63342/api/file?');

            // Add the project path so PHPstorm knows which window to make active
            href = href + '&project={{ $projectPath }}';
            $link.attr('href', href);

            // Send as an XHR request, so we don't redirect the user or have to open a new window
            $link.on('click', function(e) {
                e.preventDefault();

                $.get(href);
            })
        })
    }

    $(document).ready(function () {
        substituteLinks()

        const callback = function(mutationsList, observer) {
            for(let mutation of mutationsList) {
                if (mutation.type === 'attributes' && mutation.attributeName === 'href') {
                    substituteLinks()
                }
            }
        };

        const observer = new MutationObserver(callback);
        observer.observe(document.querySelector('.stack-main'), { attributes: true, childList: true, subtree: true });
    });
</script>

```

freekmurze:
Thanks for your work on this, I'm going to leave things as they are. Hopefully JetBrains will introduce a canonical way to handle this on all platforms soon.

reppair:
If that's the case I should we should remove the jQuery and work it with vanilla JS instead :D will add an update when have the time :)

zschuessler:
The comment here should be the way to do it cross-platform, the API Jetbrains offer is quite good:
https://github.com/facade/ignition/pull/112#issuecomment-562872875

As long as Ignition offered the ability to extend a handler, or writing one for the Jetbrains API, us non-Mac users would be happy :) 

Would you be open to the idea of a PR for either of those options?

freekmurze:
@zschuessler send a PR for this!

bleuscyther:
Here Is a version with regular JS of the updated fix provided in https://github.com/facade/ignition/pull/112#issuecomment-634488618


```PHP
<?php
/**
 * Fixes phpstorm:// protocol links Ignition uses, which don't work on Linux or Windows
 */
$projectPath = base_path();
?>

<script>

    function substituteLinks() {
        // Get all phpstorm:// protocol links
        const links = document.querySelectorAll('a[href^=phpstorm]');

        console.log(links.length);

        [...links].forEach((value, key) => {
            const link = value;
            let href = link.getAttribute('href');

            // Drop the protocol, which doesnt work on linux or windows
            href = href.replace('phpstorm://open?', 'http://localhost:63342/api/file?');

            // Add the project path so PHPstorm knows which window to make active
            href = href + '&project={{ $projectPath }}';
            link.setAttribute('href', href);

            // Send as an XHR request, so we don't redirect the user or have to open a new window
            link.addEventListener('click', (e) => {
                e.preventDefault();
                fetch(href)
            });
        })
    }


    document.addEventListener("DOMContentLoaded", function (event) {
        substituteLinks()

        const callback = (mutationsList, observer) => {
            for (let mutation of mutationsList) {
                if (mutation.type === 'attributes' && mutation.attributeName === 'href') {
                    substituteLinks()
                }
            }
        };

        const observer = new MutationObserver(callback);
        observer.observe(document.querySelector('.stack-main'), {attributes: true, childList: true, subtree: true});
    })


</script>


````

**Comment by bleuscyther** [comment]
I was trying to make a pull request by changing the file *resources/js/components/Shared/editorUrl.js*

```javascript

const editors = {
...
phpstorm:  'jetbrains://php-storm/navigate/reference?project=<projectName>&path=%path:%line':

...
}
```
In order for this to work well, we would need a way to have the project name in advance.

For example, add a new parameter **IGNITION_PHP_STORM_APP_NAME**, if the dev is using his browser on a Linux or Widows We show a message asking to set the value the **.env** file.

We could use the root dir name but it is not always the same name as the project  especially in WSL environment


```

**Comment by bleuscyther** [comment]
## Another version with the official  JetBrains way. 

Based on [their documentation on InteliJ Projects](https://www.jetbrains.com/help/idea/creating-and-managing-projects.html#project-formats)  and [on renaming projects](https://www.jetbrains.com/help/phpstorm/renaming-projects.html) it is safe to say that the project name will always be the name of the `*.iml` file. 

### New format
image
It's inside the `.idea `directory

### Old Format
image
It's a direct children of the base_path

Changing the project's name from the IDE will also change the `.iml` file name.

Here is an example of the previous "fix" where a function is used to retrieve that name:

```PHP

<?php
/**
 * Fixes phpstorm:// protocol links Ignition uses, which don't work on Linux or Windows
 */
$projectPath = base_path();

/**
 * 💥 Get the name of the project from a file in the provided directory
 * @param $path
 *
 * @return false|string
 */
function getProjectNameInDir( $path ) {
    foreach ( scandir( $path ) as $file ) {

        if ( $file == '.' || $file == '..' )
            continue;

        $dir = $path . DIRECTORY_SEPARATOR . $file;

        if ( is_dir( $dir ) ) {
            continue;
        } else {
            if ( strstr( $dir, ".iml" ) ) {
                return pathinfo( $dir, PATHINFO_FILENAME );
            }
        }
    }

    return false;
}

$projectName = '';

$iml_dirs = [
    $projectPath . DIRECTORY_SEPARATOR . '.idea',  // new project format
    $projectPath // old project format ⚠ TODO Check if 'phpstorm://open?file' works on old PhpStrom without Toolbox
];

foreach (  $iml_dirs as $path ) {

    $projectName = getProjectNameInDir( $path );
    if ( $projectName ) {
        break;
    }
}

?>

<script>

    function substituteLinks() {
        // Get all phpstorm:// protocol links
        const links = document.querySelectorAll('a[href^=phpstorm]');

        console.log(links.length);

        [...links].forEach((value, key) => {
            const link = value;
            let href = link.getAttribute('href') || '';

            //Drop the protocol, which doesnt work on linux or windows
            @if($projectName)

            // Use protocole ✅
            const urlSearchParams = new URLSearchParams(href);
            const params = Object.fromEntries(urlSearchParams.entries());
            const path = params.file || params['phpstorm://open?file']; // on some browsers the url is not parsed correctly
            const line = params.line;
            const projectName = '{{$projectName}}';
            href = `jetbrains://php-storm/navigate/reference?project=${projectName}&path=${path}:${line}`;

            @else
            // fallback
            // Use JetBrain's file API ✅
            href = href.replace('phpstorm://open?', 'http://localhost:63342/api/file?');
            console.log('☺☺');
            // Send as an XHR request, so we don't redirect the user or have to open a new window
            link.addEventListener('click', (e) => {
                e.preventDefault();
                fetch(href)
            });
            @endif

            // Add the project path so PHPstorm knows which window to make active
            {{--href = href + '&project={{ $projectPath }}';--}}
            link.setAttribute('href', href);

        })
    }


    document.addEventListener("DOMContentLoaded", function (event) {
        substituteLinks()

        const callback = (mutationsList, observer) => {
            for (let mutation of mutationsList) {
                if (mutation.type === 'attributes' && mutation.attributeName === 'href') {
                    substituteLinks()
                }
            }
        };

        const observer = new MutationObserver(callback);
        observer.observe(document.querySelector('.stack-main'), {attributes: true, childList: true, subtree: true});
    })


</script>


```


I will propose a better solution on a pull request so future installs don't need that "patch". It should involve these files if i am not mistaken:

- resources/js/components/Shared/editorUrl.js
- src/IgnitionServiceProvider.php

PR #113, opened 2019-09-06
Add RouteNotDefined Solution Provider

Te7a-Houdini:
this **PR** adds a new solution provider for Route Not Defined exceptions. for example when misspilling  `route('posts.something')`

Screenshot from 2019-09-06 02-59-59


freekmurze:
Nice work! Got some small nitpicks, which I'll take care off myself after merging.

PR #119, opened 2019-09-07
Add redis client was not installed sollution

Treggats:
In response to my own experience with a redis client mismatch between config and what was installed and also in response to this [tweet](https://twitter.com/freekmurze/status/1170080942716792832) which describes the same.

This PR adds a solution for the redis client mismatch between the config and what is installed.
It does so by checking the config for the redis client which is used and if that client is present in the vendor directory.


freekmurze:
Thank you for this PR, but we're not going to merge it.

We think that solutions for very specific errors should be in Ignition. Instead, we think it's better that the live where the actual exception is thrown.

In this case, it would be better if the exception being thrown [here](https://github.com/laravel/framework/blob/e04a7ffba8b80b934506783a7d0a161dd52eb2ef/src/Illuminate/Redis/Connectors/PhpRedisConnector.php#L73) would contain the solution as well.

Might be something PRable to Laravel.

Treggats:
Hey Freek thanks for your feedback! I'll look into the link you provided.

Issue #71, opened 2019-09-03
Provide a solution for "It is unsafe to run Dusk in production."


sebastiandedeyne:
It could suggest to install deps with the `--no-dev` flag, or modify `APP_ENV`.

Onixmiha:
P1014981624

Treggats:
On it :)

ishendyweb:
Is this issue resolved?

freekmurze:
Fixed by #121

PR #121, opened 2019-09-07
added solution for running Laravel Dusk in production

Treggats:
As requested in issue #71 

This PR provides two solutions for when Laravel Dusk is running in production.

1. install composer dependencies with the `--no-dev` flag
2. change the `APP_ENV` value


mpociot:
Thank you for the PR - can you also add a test to verify that the given exception returns those two solutions?

Treggats:
@mpociot done, added test which asserts the two solutions

mpociot:
Great job! Thank you

Issue #79, opened 2019-09-04
Add better error messages for missing validation rules.


hailwood:
Right now if we use an invalid validation rule we get something that is helpful enough.
![image](https://user-images.githubusercontent.com/709773/64217354-31e51080-cf11-11e9-88a9-3c8ab561ca1e.png)

We can do better
![image](https://user-images.githubusercontent.com/709773/64217548-ed0da980-cf11-11e9-92fd-24c6c0491a1a.png)

This should take into account any of the `validateX` functions on the Validator class and any array keys in the `$extensions` public property.

krisnaw:
Hey, if not in rush, I'd like to take this 😀

freekmurze:
This has been taken care off.

PR #125, opened 2019-09-09
#79 Add better error messages for missing validation rules.

mikaelpopowicz:


src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -0,0 +1,108 @@
...
+{
+    protected const REGEX = '/([a-zA-Z\\\\]+)::([a-zA-Z]+)/m';
+
+    public function canSolve(Throwable $throwable): bool
+    {
+        if (! $throwable instanceof BadMethodCallException) {
+            return false;
+        }
+
+        if (is_null($this->getClassAndMethodFromExceptionMessage($throwable->getMessage()))) {
```
Gummibeer: Couldn't this get simplified by a simple `return is_null()`? Or did you do this for readability?

src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -0,0 +1,108 @@
...
+
+    public function getSolutions(Throwable $throwable): array
+    {
+        return [
+            BaseSolution::create('Unknown Validation Rule')
+                ->setSolutionDescription($this->getSolutionDescription($throwable)),
+        ];
+    }
+
+    public function getSolutionDescription(Throwable $throwable): string
```
Gummibeer: Shouldn't it be protected?

src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        if (! $this->canSolve($throwable)) {
```
Gummibeer: If it's protected this condition could get removed.

src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -0,0 +1,108 @@
...
+            return '';
+        }
+
+        extract($this->getClassAndMethodFromExceptionMessage($throwable->getMessage()), EXTR_OVERWRITE);
+
+        $possibleMethod = $this->findPossibleMethod($class, $method);
+        $rule           = lcfirst(str_replace('validate', '', $possibleMethod));
```
Gummibeer: Indentation of the assignment isn't part of the style guide!?

src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -0,0 +1,108 @@
...
+        ];
+    }
+
+    public function getSolutionDescription(Throwable $throwable): string
+    {
+        if (! $this->canSolve($throwable)) {
+            return '';
+        }
+
+        extract($this->getClassAndMethodFromExceptionMessage($throwable->getMessage()), EXTR_OVERWRITE);
```
Gummibeer: A simple list assignment should be enough and is more common.

src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -0,0 +1,108 @@
...
+        return "Did you mean `{$rule}` ?";
+    }
+
+    protected function getClassAndMethodFromExceptionMessage(string $message): ?array
+    {
+        if (! preg_match(self::REGEX, $message, $matches)) {
+            return null;
+        }
+
+        if ($matches[1] !== Validator::class) {
```
Gummibeer: If you only want `Validator` you could adjust the regex?

src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -0,0 +1,108 @@
...
+            return null;
+        }
+
+        if (! Str::startsWith($matches[2], 'validate')) {
```
Gummibeer: Also this could be moved into regex.

src/SolutionProviders/UnknownValidationSolutionProvider.php
mikaelpopowicz: The solution provider is a copy of BadMethodCallSolutionProvider.

src/SolutionProviders/UnknownValidationSolutionProvider.php
mikaelpopowicz: I don't know

src/SolutionProviders/UnknownValidationSolutionProvider.php
mikaelpopowicz: Yes I can

src/SolutionProviders/UnknownValidationSolutionProvider.php
mikaelpopowicz: Yes

src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -37,53 +32,38 @@ public function getSolutions(Throwable $throwable): array
...
-            return '';
-        }
+        $method = $this->getMethodFromExceptionMessage($throwable->getMessage());
 
-        extract($this->getClassAndMethodFromExceptionMessage($throwable->getMessage()), EXTR_OVERWRITE);
-
-        $possibleMethod = $this->findPossibleMethod($class, $method);
-        $rule           = lcfirst(str_replace('validate', '', $possibleMethod));
+        $possibleMethod = $this->findPossibleMethod($method);
+        $rule = lcfirst(str_replace('validate', '', $possibleMethod));
```
Gummibeer: Isn't it a `Str::slug()` instead of a simple `lcfirst()`? At the end this is used by the validator.

mikaelpopowicz: Actually it uses Str::snake

src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -0,0 +1,84 @@
...
+
+                return $percentage;
+            })->first();
+    }
+
+    protected function getAvailableMethods(): Collection
+    {
+        $class = new ReflectionClass(Validator::class);
+
+        $extensions = Collection::make((\Illuminate\Support\Facades\Validator::make([], []))->extensions)
```
freekmurze: Can we do this without making use of the facade?

tests/Solutions/UnknownValidationSolutionProviderTest.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    protected function getBadMethodCallException(string $rule = 'number'): \BadMethodCallException
+    {
+        $default = new \BadMethodCallException('Not a validation rule exception!');
+        try {
+            $validator = Validator::make(['number' => 10], ['number' => "{$rule}"]);
+            $validator->validate();
+            return $default;
+        } catch (\BadMethodCallException $badMethodCallException) {
```
freekmurze: Nitpick: import these

tests/Solutions/UnknownValidationSolutionProviderTest.php
```diff
@@ -0,0 +1,74 @@
...
+            return $badMethodCallException;
+        } catch (\Exception $exception) {
+            return $default;
+        }
+    }
+
+    public function rulesProvider()
```
freekmurze: Add a return type hint

src/SolutionProviders/UnknownValidationSolutionProvider.php
Gummibeer: Should be possible via the factory service where the extensions are original registered.

mikaelpopowicz: We may use `app('validator')->make([], [])`

mikaelpopowicz: The factory service `$extensions` property is protected.

Gummibeer: This could be solved by a reflection. But this is kind of philosophic if a reflection is better than a make method call.

src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -0,0 +1,84 @@
...
+    protected function getMethodFromExceptionMessage(string $message): ?string
+    {
+        if (! preg_match(self::REGEX, $message, $matches)) {
+            return null;
+        }
+
+        return $matches['method'];
+    }
+
+    protected function findPossibleMethod(string $invalidMethodName)
```
hailwood: This method could probably (and likely should for consistency) be replaced to use the [Facade\Ignition\Support\StringComparator](https://github.com/facade/ignition/blob/da28c127c38e6d7afe6b9d2c687c6b15277d5e4d/src/Support/StringComparator.php) class.

See the Route Not Defined solution provider for context. https://github.com/facade/ignition/blob/da28c127c38e6d7afe6b9d2c687c6b15277d5e4d/src/SolutionProviders/RouteNotDefinedSolutionProvider.php#L50

src/SolutionProviders/UnknownValidationSolutionProvider.php
mikaelpopowicz: @freekmurze what option do you prefer ?

src/SolutionProviders/UnknownValidationSolutionProvider.php
mikaelpopowicz: I just tried. With the wrong rule `unik` i got :

- `unique` with similar_text
- `uuid` with StringComparator::findClosestMatch

freekmurze: Yeah, that's a good suggestion @Gummibeer, let's use that.

mikaelpopowicz: @freekmurze may we just add a new static method `findSimilarText(array $strings, string $input): array` in StringComparator ?

freekmurze: I would be ok with that.

src/SolutionProviders/UnknownValidationSolutionProvider.php
```diff
@@ -37,7 +38,15 @@ protected function getSolutionDescription(Throwable $throwable): string
     {
         $method = $this->getMethodFromExceptionMessage($throwable->getMessage());
 
-        $possibleMethod = $this->findPossibleMethod($method);
+        $possibleMethod = StringComparator::findSimilarText(
+            $this->getAvailableMethods()->toArray(),
+            $method
+        );
+
+        if (empty($possibleMethod)) {
```
Gummibeer: Personally I prefer an explicit `null` check - `is_null()` or `=== null`. `empty()` is too broad and the explicit null check improves readability.

freekmurze:
Thanks!

PR #129, opened 2019-09-10
Add syntax highlighting

mpociot:
This PR adds syntax highlighting to our code snippet.

If the `auto` theme is selected, we will use a media query listener to detect if we want to use the light or the dark color scheme.

Light:
image

Dark:
image


resources/js/components/Stack/Snippet.vue
```diff
@@ -72,10 +79,32 @@ export default {
...
     data() {
         return {
+            detectedTheme: mediaQuery.matches ? 'dark' : 'light',
             firstSelectedLineNumber: null,
         };
     },
 
+    watch: {
+        selectedFrame: value => {
+            highlightState = null;
```
sebastiandedeyne: After the previous change, this block should be:

```
selectedFrame() {
    this.highlightState = null;
}
```

sebastiandedeyne: It's easy to forget `this` every now and then when you've been working in React for months :D

resources/js/components/Stack/Snippet.vue
```diff
@@ -35,27 +35,34 @@
...
+hljs.registerLanguage('php', require('highlight.js/lib/languages/php'));
+
 import ExceptionClass from '../Shared/ExceptionClass.vue';
 import FilePath from '../Shared/FilePath.vue';
 import LineNumber from '../Shared/LineNumber.vue';
 import editorUrl from '../Shared/editorUrl';
 
+let highlightState = null;
+
+const mediaQuery = matchMedia('(prefers-color-scheme: dark)');
```
sebastiandedeyne: This should be moved inside `data` instead of outside of the component

resources/js/components/Stack/Snippet.vue
```diff
@@ -35,27 +35,34 @@
...
 <script>
+import hljs from 'highlight.js/lib/highlight';
+hljs.registerLanguage('php', require('highlight.js/lib/languages/php'));
+
 import ExceptionClass from '../Shared/ExceptionClass.vue';
 import FilePath from '../Shared/FilePath.vue';
 import LineNumber from '../Shared/LineNumber.vue';
 import editorUrl from '../Shared/editorUrl';
 
+let highlightState = null;
```
sebastiandedeyne: This should be moved to the component's `create` hook.

```
created() {
    this.highlightState = null;
}
```

PR #139, opened 2019-09-12
protect post routes with middleware

Gummibeer:
Like discussed with @freekmurze this adds a little protection to the two post routes. In extension of #93


freekmurze:
Only thing missing here seems to be a test, could you add that @Gummibeer ?

Gummibeer:
Unit or feature?

freekmurze:
Depends what you understand under those terms. I would have a test that sets up a route with the middleware applied to it.

The test would test just make calls to that route.

Gummibeer:
Ok, so feature/implementation. Unittest, for me, would be calling the middleware handle method directly and expect any result.
Will do it tomorrow.

freekmurze:
👍

Gummibeer:
unittests added in https://github.com/facade/ignition/pull/139/commits/c5ded3ea40c9961e87cabc89775c5a2390a51fae

freekmurze:
Thanks for your work on this. 

@mpociot if you agree with these changes too, we can merge them.

mpociot:
Thank you for this! :)

PR #142, opened 2019-09-13
Add no-index on error page

willemvb:
Prevent indexing for errors that are visible on badly configured public projects.


freekmurze:
Thanks!

PR #145, opened 2019-09-15
Adds margin on bottom of error card to give breathing room

awebartisan:
This pull request adds margin on bottom to give breathing room, and also, to let us know that this end of page.

### DIFF
--- a/resources/js/components/App.vue
+++ b/resources/js/components/App.vue
@@ -1,7 +1,7 @@
 <template>
     <div>
         <Summary />
-        <div class="layout-col mt-12">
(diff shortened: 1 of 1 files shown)


willemvb:
The margin-less bottom is actually on purpose.
The idea behind it is that you can quickly scroll down without too much precision and always have a minimum of 100vh screen estate at your disposal for all tabbed content.

This PR breaks this behaviour, eg. in an empty debug tab where the top filters don't align anymore when you've scrolled to the bottom completely.

image

For now, we think the scrollbars provide you enough information that you reached the end of the page.

PR #147, opened 2019-09-16
TestCommand: Prevent undefined index exception

dennisoderwald:
If `channels` is not set, an exception will occur: "Undefined index: channels".

With other logging methods such as stderr, for example, a "channels" array does not exist. This fix fixes this.


freekmurze:
Thanks!

PR #154, opened 2019-09-18
Added url to src of script tag

guhweb:
This modification is backwards compatible, and this setting allows the component to be used by those not using the built-in php server, such as XAMPP.
The project url in XAMPP is something like http://localhost/laravel-instalation/public, so the path / <?= $ housekeepingEndpoint ?>/assets does not understand this component.

### DIFF
--- a/resources/views/errorPage.php
+++ b/resources/views/errorPage.php
@@ -22,7 +22,7 @@
 
     <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.7.0/css/all.css">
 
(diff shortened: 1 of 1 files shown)


freekmurze:
If the config is cached, the env variable will not return anything. It's better to use the config helper.

guhweb:
@freekmurze check it out! True, it just hit me this situation ... Want to remove this pull request and I send a new one?

freekmurze:
Feel free to modify this PR.

guhweb:
Sended in Pull #155. Tks @freekmurze

PR #155, opened 2019-09-18
env() to config() helper switch

guhweb:
Changed to use helper to avoid problems with setting cache.


mpociot:
Thank you for the PR. I noticed that the error also appeared for the custom tab css and javascript files. So I had to fix it in multiple locations. I fixed this in this commit https://github.com/facade/ignition/commit/622de16dad9d2d8ecb1585db8a942ed918d8f430 and tagged version 1.7.1.

PR #132, opened 2019-09-10
Resolve configuration from the injected application instead of the helper

sebdesign:
Fixes #131

Issue #131, opened 2019-09-10
Target class [config] does not exist


sebdesign:
PHP: 7.3.5
Laravel: 6.0.1
Ignition: 1.6.0

Hello,

I just encountered a weird bug while running some unit tests.

When dump any value, using `dump()`, or `Collection::dump()`,
and then I perform any database query, the `QueryRecorder` throws an exception while it tries to get the configuration.

Here is a minimal test case that replicates the problem:

```php
<?php

namespace Tests\Feature;

use App\User;
use Tests\TestCase;

class ExampleTest extends TestCase
{
    /**
     * A basic feature test example.
     *
     * @return void
     */
    public function testExample()
    {
        // Dump any value
        dump(1);

        // Then perform a query
        User::first();
    }
}
```

This is the stack trace:

```
1) Tests\Feature\ExampleTest::testExample
Illuminate\Contracts\Container\BindingResolutionException: Target class [config] does not exist.

/example/vendor/laravel/framework/src/Illuminate/Container/Container.php:806
/example/vendor/laravel/framework/src/Illuminate/Container/Container.php:682
/example/vendor/laravel/framework/src/Illuminate/Container/Container.php:630
/example/vendor/laravel/framework/src/Illuminate/Foundation/Application.php:768
/example/vendor/laravel/framework/src/Illuminate/Foundation/helpers.php:121
/example/vendor/laravel/framework/src/Illuminate/Foundation/helpers.php:283
/example/vendor/facade/ignition/src/QueryRecorder/QueryRecorder.php:30
/example/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php:347
/example/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php:196
/example/vendor/laravel/framework/src/Illuminate/Database/Connection.php:829
/example/vendor/laravel/framework/src/Illuminate/Database/Connection.php:683
/example/vendor/laravel/framework/src/Illuminate/Database/Connection.php:636
/example/vendor/laravel/framework/src/Illuminate/Database/Connection.php:334
/example/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php:2140
/example/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php:2128
/example/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php:2572
/example/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php:2129
[...]

PR #168, opened 2019-09-26
Resolve configuration from the injected app instead of the helper

gdebrauwer:
#131 was initially fixed by #132, but the issue has been reintroduced.
This pull request fixes it again 🙃


freekmurze:
Thanks!

PR #174, opened 2019-10-01
Change how shell Process instance is created while adding Git information

### LINKED ISSUE #173: Use of deprecated way for shell process instantiation in AddGitInformation middleware
Symfony's `Process` class used for adding GIT information (`AddGitInformation` middleware) has this piece of code:
```
    if (!\is_array($command)) {
        @trigger_error(sprintf('Passing a command as string when creating a "%s" instance is deprecated since Symfony 4.2, pass it as an array of its arguments instead, or use the "Process::fromShellCommandline()" constructor if you need features provided by the shell.', __CLASS__), E_USER_DEPRECATED);
    }
```
As you can see, it triggers an error when command used during `Process` instantiation (new `Process(...)`) is a string. Because the error is suppressed, nothing will break, however defining command as a string is deprecated.

There is a simple fix for that which I'm happy to contribute.

Regards

golonix:
Resolves #173

### DIFF
--- a/src/Middleware/AddGitInformation.php
+++ b/src/Middleware/AddGitInformation.php
@@ -47,7 +47,7 @@ class AddGitInformation
 
     protected function command($command)
     {
-        $process = new Process($command, base_path());
+        $process = Process::fromShellCommandline($command, base_path());
 
         $process->run();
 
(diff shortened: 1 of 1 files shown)


mpociot:
Thank you for the PR @golonix !

PR #175, opened 2019-10-01
Add make:solution command

### LINKED ISSUE #135: Suggestion: Provide a make command
Maybe it's a good idea if there will be an artisan command to make custom solutions.

So something like: `php artisan make:solution`.

If you have any questions feel free to ask.

Te7a-Houdini:
this **PR** resolves #135 . and adds ``` php aritsan make:solution ExampleSolution```


src/Commands/SolutionMakeCommand.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Get the stub file for the generator.
+     *
+     * @return string
+     */
+    protected function getStub()
+    {
+        return $this->option('runnable')
+            ? __DIR__.'/stubs/solution-runnable.stub'
```
freekmurze: Rename this file to `runnable-solution.stub`

Te7a-Houdini: @freekmurze Done

src/Commands/SolutionMakeCommand.php
```diff
@@ -0,0 +1,65 @@
...
+     * @var string
+     */
+    protected $name = 'make:solution';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Create a new custom solution class';
```
mpociot: Can you change the description to "Create a new custom Ignition solution class"?
Since the make command is not prefixed, it would be good to have the description as descriptive as possible.

Te7a-Houdini: @mpociot done

mpociot:
Thank you!

PR #182, opened 2019-10-05
Add conditional line number

### LINKED ISSUE #26: Show line number with dumps
Would be nice if the line number was shown with the dumps. Sometimes the same text is dumped a couple of times in the same file.

image

jasonvarga:
Adds the line number to the `FilePath` component. The prop was already there and being passed in, it was just not being rendered.

This closes #26, but would also make the number visible in the path above the snippet. You'd get a double up between the line number badge and the line number in the path. Not sure if that's a problem.

image

If it is, maybe an `includeLineNumber` prop could be introduced.


freekmurze:
Thanks!

PR #184, opened 2019-10-07
Removing package-lock.json

justinasposiunas:
I suppose that yarn is preferred package manager over npm in this project because of a few reasons:

1. It is being used there:

https://github.com/facade/ignition/blob/9b312a29fb6c781d5e2f4e41fb42ce3fde63a237/package.json#L43-L53

2. `yarn.lock` is added earlier and it is being updated. `package-lock.json` appeared in repo after merging this PR: https://github.com/facade/ignition/pull/47/

The presence of that file might confuse other contributors. So probably it is better to remove `package-lock.json` instead of working on automatic synchronization with `yarn.lock` and supporting both package managers properly.


freekmurze:
Meanwhile, I've removed an ignore  `package-lock.json`. We haven't heard complaints on us using `yarn` from other contributors yet, so probably we're going to leave the yarn specific stuff in there for now.

PR #194, opened 2019-10-11
Allow empty solution titles (fixes #172)

### LINKED ISSUE #172: Allow solutions without titles
Some exception messages are already pretty user friendly. In these cases, the solution title just ends up repeating the message.

You can currently pass in an empty string (`BaseSolution::create('')`), but the spacing ends up a little odd because there's an empty `h2`.

willemvb:
Don't render `h2` if solution title is an empty string


freekmurze:
Added to v2 of ignition as v2.0.8

PR #195, opened 2019-10-11
Add ignition:make:solution-provider command

Te7a-Houdini:
this **PR** adds php aritsan ignition:make:solution-provider ExampleSolutionProvider


freekmurze:
Thank you for this, I've added it to v2.

PR #197, opened 2019-10-13
Name routes

ankurk91:
Hi,

It is good to have named routes, this is very useful when we need to whitelist/blacklist routes in packages like [Ziggy](https://github.com/tightenco/ziggy).


freekmurze:
Could you prefix the names with `ignition.` so that there aren't any collisions with existing route names in apps where Ignition will be installed?

ankurk91:
@freekmurze 
I have already applied
` 'as' => 'ignition.',` on group

freekmurze:
Oh, I missed that, thanks!

PR #198, opened 2019-10-13
Simplify default Laravel installation

mpociot:
Default Laravel installations now automatically see the `flare:test` command and the `ignition:make:*` command. In addition to that, a `route:list` shows the Ignition routes.

This PR only shows the flare:test command when a Flare key is configured.
The ignition:* commands have to explicitly be enabled
The routes don't get registered on the CLI to keep the route:list clean.


freekmurze:
Perfect!

PR #200, opened 2019-10-14
Add solution provider for missing mix manifest

travisobregon:
I'm not sure if you think that there should be a title or not, but I left it empty assuming #194 is merged.

### DIFF
--- a/src/SolutionProviders/MissingMixManifestSolutionProvider.php
+++ b/src/SolutionProviders/MissingMixManifestSolutionProvider.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Facade\Ignition\SolutionProviders;
+
+use Facade\Ignition\Exceptions\ViewException;
+use Illuminate\Support\Str;
+use Throwable;
+use Facade\IgnitionContracts\BaseSolution;
+use Facade\IgnitionContracts\HasSolutionsForThrowable;
+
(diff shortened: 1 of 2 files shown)


src/SolutionProviders/MissingMixManifestSolutionProvider.php
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        return Str::startsWith($throwable->getMessage(), 'The Mix manifest does not exist');
+    }
+
+    public function getSolutions(Throwable $throwable): array
+    {
+        return [
+            BaseSolution::create('')
+                ->setSolutionDescription('Generate the Mix manifest by using `npm install && npm run dev`.')
```
freekmurze: I would not use `npm` in the description, because people might use `yarn`. Instead I would use a more generic message like: "You shoud regenerate your assets", and add a link to the docs: https://laravel.com/docs/6.0/mix#running-mix

AdrianMrn: @freekmurze we could check if a `package.lock` or `yarn.lock` file exists in the project root, although this will not catch 100% of cases (some people have both, even though `yarn` shouts warnings about this).

This could also be a runnable solution if we do the check. Although, it will only work if users have a `dev` script registered in their `package.json`, so I'm not entirely convinced myself.

freekmurze:
Thank you, but I'm going to pass on this one for now.

PR #208, opened 2019-10-21
Don't rely on external resources

### LINKED ISSUE #134: Suggestion: do not load external resources
I hope I don't sound accusatory as english is not my main language.

I believe that it would be better for a development focused package to not load any external resources or make any unsolicited calls on the backend side. I have not reviewed the codebase but I assume that the latter is also true, as long as you don't define a Flare API key.

The external resource in question is the FontAwesome package:
https://github.com/facade/ignition/blob/107bc007da8268bffa6a6337294095ebc27ecb48/resources/views/errorPage.php#L22

Could you instead bundle this by default so no external connections are made? Since at least in my case the browser runs with very relaxed settings for local domains.

willemvb:
This PR should fix #134.
 
- Remove link to Font Awesome CDN
- Set icons as SVG Vue components
- Use SVG symbols to reduce duplication in output (especially the 'pencil' icon)


resources/js/components/Icons/ArrowDownIcon.vue
```diff
@@ -0,0 +1,5 @@
+<template>
```
sebastiandedeyne: Since these have been refactored to `use` tags, are separate icon components worth it over a generic `<Icon name="arrow-down" />`?

willemvb: Good point, rewrote to a generic `Icon` component. Can you have an extra look?

miclf:
Hello there! I’m wondering if there is any plan to merge this pull request anytime soon? :)

I recently discovered that one of the slightly annoying effects of the issue being solved here is that, when you develop while being offline (e.g. in NMBS/SNCB trains), the links to open code editors are not even displayed, since the icons are not downloaded and there isn’t any fallback text as link content.
The links are obviously present in the HTML page, but they have dimensions of zero.

**Illustration with Firefox devtools:**

Capture d’écran 2019-12-23 à 02 44 42

PR #212, opened 2019-10-26
Add indicator that solution is running

mbardelmeijer:
When using long-running solutions, there is currently no indication that the solution is running in the background.

Example: asset is missing, solution is to compile assets, and that takes a few seconds

This ensures that the user knows that the solution is running. Also prevents solutions to be run multiple times by pressing the button multiple times.

I'm no vue expert so hopefully the implementation is alright, let me know if i need to change anything :)

### DIFF
--- a/resources/js/components/Solutions/SolutionCard.vue
+++ b/resources/js/components/Solutions/SolutionCard.vue
@@ -60,10 +60,12 @@
                                             canExecuteSolutions === true &&
                                             executionSuccessful === null
                                     "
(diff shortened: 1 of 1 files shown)


fgilio:
This looks like a good thing to have built in, any issue preventing the merge?

freekmurze:
Ok for me. 

@sebastiandedeyne can this get your blessing too?

sebastiandedeyne:
Looks good to me!

freekmurze:
Thanks!

PR #215, opened 2019-11-04
Ignore JS tests

### LINKED ISSUE #214: Tests being deployed with package
I am using laravel. So I found these tests that are being downloaded with the `ignition` package.

The path of the **tests** are `vendor/facade/ignition/resources/js/stack/__tests__`.

They are just the only tests there in laravel .. are they meant to be shipped with the package?
I am new to this stuff, so just wanted to report it.

JayBizzle:
Fixes #214

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -13,3 +13,4 @@
 /yarn.lock          export-ignore
 /package-lock.json  export-ignore
 /CONTRIBUTING.md    export-ignore
+__tests__           export-ignore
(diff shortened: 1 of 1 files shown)


sebastiandedeyne:
Maybe we should ignore entire `resources/js` and `resources/css` instead?

freekmurze:
I think it's safe to ignore those directories.

JayBizzle:
> Maybe we should ignore entire `resources/js` and `resources/css` instead? Ping @mpociot

Agreed, update the file 👍

mpociot:
Looks good :)

PR #216, opened 2019-11-08
Correctly detect dump location from ddd

netpok:
This pull request fixes that current the helper.php is shown when using ddd.


freekmurze:
Thanks!

PR #228, opened 2019-12-20
do not overflow the "card-details"

browner12:
when using automated Browser testing with screenshots (like Dusk), we can't see the entire name of a long View file.

switching this class forces the text to wrap, which allows us to see the whole file in the screenshot.


freekmurze:
We've decided not to merge this change, because we like the current behaviour better for humans.

PR #230, opened 2019-12-30
Add exception to the bottom of the html

SjorsO:
This PR is similar to https://github.com/facade/ignition/pull/206

I just ran into a test that failed because somewhere in a controller it fails a safety check and calls `abort(422)`. The test fails with the following message: "Response status code [422] is not a redirect status code.". Because this message isn't very useful, i decide to `dump()` the `TestResponse`. Doing that currently shows something like this:

image

The exception message is at the top of the html, which means I have to scroll through a lot of inline javascript before seeing useful information about the error.

This PR adds the exception message to the bottom of the html, so that dumping the `TestResponse` immediately shows useful information:

image

### DIFF
--- a/resources/views/errorPage.php
+++ b/resources/views/errorPage.php
@@ -57,5 +57,8 @@
 <script>
     Ignition.start();
 </script>
+<!--
+<?=$throwableString?>
+-->
 </body>
 </html>


freekmurze:
Thanks!

PR #243, opened 2020-01-30
Change DumpRecorder dependency to use contract instead of Application.

filipac:
I've been trying to add Ignition to an OctoberCms application and so far the only thing that stops this is the hard-coded dependency to Illuminate\Foundation\Application instead of Illuminate\Contracts\Foundation\Application - because October changes the application instance with a custom one that implements the Contract.

### DIFF
--- a/src/DumpRecorder/DumpRecorder.php
+++ b/src/DumpRecorder/DumpRecorder.php
@@ -2,7 +2,7 @@
 
 namespace Facade\Ignition\DumpRecorder;
 
-use Illuminate\Foundation\Application;
+use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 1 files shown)


freekmurze:
Seems good to me. Do you see a problem with this @mpociot?

mpociot:
Looks good 👍

freekmurze:
Thanks!

LukeTowers:
@filipac @freekmurze @mpociot we're actually working on an official plugin to integrate with Ignition right now (https://github.com/rainlab/ignition-plugin) and we've fixed this in the core October build (https://github.com/octobercms/library/commit/e33163d5b288f05ffec66ab9e55ee96c1457164b), since 2.0 isn't usable for October CMS as we only ever follow LTS releases of Laravel.

PR #248, opened 2020-02-12
Extract ui

AdrianMrn:
**Shares should be tested before merging this.**

This PR extracts the UI into a separate package (hosted at `@facadecompany/ignition-ui` on npm), so the UI can be used by other error notifiers.

**TODO**: Remove the JS test from this repo's GitHub Actions, and move them to the `ignition-ui` repo.

@mpociot @freekmurze please check the following:
I changed this line: https://github.com/facade/ignition/blob/e3c630507232719cb83c3e3f00dbc1212e75f897/src/Http/Controllers/ShareReportController.php#L14

and the `ignition-ui` package now also doesn't send the report as a string, but as an object. I don't know why this was implemented originally, but it doesn't seem to be causing any issues now. Do you think this could be a problem?

Read here for more information on why I did this: https://github.com/facade/ignition-js/issues/13


freekmurze:
We'll hold off on this for now.

PR #255, opened 2020-03-02
Added Laravel 7 Support

RahulDey12:
Because of this, I am not able to upgrade laravel 7


freekmurze:
We have our separate branch ready, which adds compatibility with Laravel 7. We'll tag it soon.

PR #260, opened 2020-03-06
Add a solution for unsupported cache::lock() driver

SjorsO:
Not every cache driver supports using `Cache::lock()`. This PR adds a solution for the error you get if you try using atomic locks when using an unsupported driver.

[Cache::lock() documentation](https://laravel.com/docs/7.x/cache#atomic-locks)

---
2020-05-02 edit: The database driver will soon support atomic locks: https://github.com/laravel/framework/pull/32639

---

### Screenshot of the new solution
image


SjorsO:
Closing this because most of the common cache drivers now support atomic locks.

PR #264, opened 2020-03-16
Added "Undefined Property" solution

musa11971:
This solution provides the developer with a suggested object property name, if a similar one is found. This is useful in the case of a typo.  
You could compare this solution to [UndefinedVariableSolutionProvider](https://github.com/facade/ignition/blob/master/src/SolutionProviders/UndefinedVariableSolutionProvider.php), but for object properties.

## Example
Consider the following class:
```php
class Car 
{
	public $name;
}
```

Scenarios:
| Code  | Solution triggered? | Description
| ------------- | ------------- | ------------- |
| `$car->na`  | ❌ No  | - |
| `$car->nam`  | ✅ Yes  | Did you mean App\Car::$name ? |
| `$car->name`  | ❌ No  | - |
| `$car->namee`  | ✅ Yes  | Did you mean App\Car::$name ? |
| `$car->nameee`  | ✅ Yes  | Did you mean App\Car::$name ? |
| `$car->nameeee`  | ❌ No  | - |
| `$car->something`  | ❌ No  | - |

## Testing
```bash
./vendor/bin/phpunit --filter=UndefinedPropertySolutionProvider
```


freekmurze:
Thank you!

PR #265, opened 2020-03-17
Fix execute solution route not definded.

nqxcode:
Fixed problem with action Facade\Ignition\Http\Controllers\ExecuteSolutionController not defined.


freekmurze:
Thanks!

PR #269, opened 2020-04-07
Support Laravel 8

driesvints:
`laravel new myapp --dev` is currently broken because Ignition doesn't has the `^8.0` constraint yet.


freekmurze:
Thanks!

PR #271, opened 2020-04-15
Allow override of Dumper via SERVER variable

Korko:
Copy behavior from Symfony VarDumper.

PHP_SAPI cannot be changed but sometimes it can be helpful to enforce the dumper.


freekmurze:
I like the added functionality, but I don't like the implementation.

Could you refactor this so we have a dedicated function to returns the right dumper. Use early returns. That will greatly improve readability.

Thanks!

freekmurze:
I'll change this myself, thanks!

PR #274, opened 2020-05-17
Handle exceptions in case the request doesn't have a user

stancl:
In my multi-tenant app, the auth guard for some requests will only work once the DB connections are switched, so retrieving the user too early -- like Ignition does -- will result in an error.

With this change, the user probably won't be available in Ignition reports on tenant routes, but at least it won't throw an exception.


freekmurze:
Thanks!

PR #279, opened 2020-06-06
Ignore files when exporting package

simonschaufi:
This commit is part of a campaign to reduce the amount of data transferred to save global bandwidth and reduce the amount of CO2. See https://github.com/Codeception/Codeception/pull/5527 for more info.

If I added to many files to the blacklist, please adjust the file yourself!


freekmurze:
Thanks!

PR #281, opened 2020-07-02
DocBlock support for BadeMethodCallSolution 

cbl:
This pull request adds methods that are defined in a classes docblock to the `BadMethodCallSolutionProvider`.

```php
/**
 * @method string myMethod()
 */
class MyClass
{

}
```

Helpfull to give better solutions for facades.

### DIFF
--- a/src/SolutionProviders/BadMethodCallSolutionProvider.php
+++ b/src/SolutionProviders/BadMethodCallSolutionProvider.php
@@ -2,13 +2,14 @@
 
 namespace Facade\Ignition\SolutionProviders;
 
-use BadMethodCallException;
-use Facade\IgnitionContracts\BaseSolution;
-use Facade\IgnitionContracts\HasSolutionsForThrowable;
-use Illuminate\Support\Collection;
+use Throwable;
 use ReflectionClass;
 use ReflectionMethod;
-use Throwable;
+use BadMethodCallException;
+use Illuminate\Support\Collection;
+use Facade\IgnitionContracts\BaseSolution;
+use phpDocumentor\Reflection\DocBlockFactory;
+use Facade\IgnitionContracts\HasSolutionsForThrowable;
 
 class BadMethodCallSolutionProvider implements HasSolutionsForThrowable
 {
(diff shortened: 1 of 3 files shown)


freekmurze:
The tests seem to be failing for this one.

src/SolutionProviders/BadMethodCallSolutionProvider.php
```diff
@@ -83,7 +83,11 @@ protected function getAvailableMethods($class): Collection
 
     public function getAvailableMethodsFromClassDoc(ReflectionClass $class): Collection
     {
-        $doc = DocBlockFactory::createInstance()->create($class->getDocComment());
+        if (! $doc = $class->getDocComment()) {
+            return Collection::make([]);
+        }
+
+        $doc = DocBlockFactory::createInstance()->create($doc);
```
cbl, line 90: Fixed, for empty doc blocks.

tests/ExceptionSolutionTest.php
```diff
@@ -124,4 +125,28 @@ public function it_can_propose_a_solution_for_missing_app_key_exceptions()
...
+
+/**
+ * @method void test()
+ */
+class ClassWithDocBlock
+{
+    public function __call($method, $parameters = [])
+    {
+        throw new BadMethodCallException(static::class.'::'.$method);
+    }
```
cbl, line 151: Added a test for doc block method solution support.

cbl:
I fixed the docblock method solution for empty docblocks and added a test @freekmurze

freekmurze:
Thanks for your work on this. I'm not going to merge it in, because we don't want any new dependancies at the moment.

I might change my stance on this in the future.

cbl:
Hey, I totally understand this, since Ignition is a dependency of Laravel and has millions of downloads.

However, the `phpdocumentor/reflection-docblock` is a dependency of Laravel anyway through `phpunit/phpunit` -> `phpspec/prophecy` -> `phpdocumentor/reflection-docblock`. If this changes your opinion on this I would offer to add support for nested methods and property suggestion from doc blocks. However, I can also understand if you don't want to add dependencies.

Issue #275, opened 2020-05-20
 DumpRecorder breaks tests with dump() & Application::getNamespace() calls in it


tzurbaev:
- Laravel version: 7.12.0
- Ignition version: 2.0.4
- Reproduction repository: https://github.com/tzurbaev/ignition-tests-dump-namespace (see tests or latest commit).
- Travis logs: https://travis-ci.org/github/tzurbaev/ignition-tests-dump-namespace/builds/689137616

This is very specific case, so I'll understand if this won't get fixed. Also, this might not be related to Ignition itself, but since I've noticed this bug with Ignition I decided to report it here first.

---

If you're using `dump` function in your tests and then in some way calling `Illuminate\Contracts\Foundation\Application::getNamespace()` method, your tests might fail because `Application` instance might be overridden at the time of the `getNamespace()` call.

1. The `getNamespace()` method expects that its instance has been created with the `$basePath` property (since it tries to read application's `composer.json` file to detect namespace in first call).
2. Laravel creates initial instance of `Application` with correct base path so `getNamespace` method works as expected. Behind the scenes `Application`'s constructor also updates singleton instance from the `Application::registerBaseBindings()` method. So `Container` singleton is being updated after every instantiation of `Application` class.
3. The `Facade\Ignition\DumpRecorder\DumpRecorder` class has dependency of `Application` class but this does not break namespace resolver itself. The `register()` method makes instance of `Facade\Ignition\DumpRecorder\DumpHandler` from the container and that's where something that I can't figure out happens.
4. The `DumpHandler` class has dependency of `DumpRecorder` class (that has dependency of `Application` class as was mentioned earlier). And it seems that in this case the `DumpRecorder` class is being created with the fresh instance of the `Application` class that has no `$basePath` property set (it also overrides `Container` singleton as well).
5. Any subsequent calls to `Application::getNamespace()` method will fail with the `ErrorException: file_get_contents(/composer.json): failed to open stream: No such file or directory` exception.

I wasn't able to trace back to the original reason why `DumpHandler -> DumpRecorder` chain creates fresh instance of `Application` instance, so I need help with this (this might be some bug in Laravel core that should be fixed in framework's codebase).

For now I have 2 suggestions that can fix this bug:

1. Create instance of `DumpHandler` manually instead of using `$this->app->make()` method:

```diff
index 027b11e..cf6d072 100644
--- a/src/DumpRecorder/DumpRecorder.php
+++ b/src/DumpRecorder/DumpRecorder.php
[...]

PR #286, opened 2020-07-13
Make DumpHandler instance manually in DumpRecorder::register()

tzurbaev:
As described in #275, `$this->app->make()` call on `DumpHandler` overrides singleton instance of `Application` class and this breaks any calls to `Application::getNamespace()` method after `dump`ing something in tests.


freekmurze:
Thanks!

Issue #272, opened 2020-04-26
QueueServiceProvider optional


SunMar:
Hi,

Today while experimenting I came across:

https://github.com/facade/ignition/blob/7ada5e279beee1a33b941d7365f5a4528bc5b197/src/IgnitionServiceProvider.php#L83

That line makes enabling the `QueueServiceProvider` mandatory. If it's disabled the `IgnitionServiceProvider` will throw an exception on that line. Does it make sense to do a `$this->app->bound('queue')` check before attempting to setup the queues? Unless you actually connect Ignition to Flare I don't see why having queues available should be mandatory. But please enlighten me if I'm missing or not understanding something :).

Since Laravel by default is a bit bloated, when I start a new project I prefer to disable all service providers and facades in `config/app.php` and only enable them on an as-needed basis. Simple applications often don't need use a queue, then keeping the service provider enabled is a bit pointless (though probably most people don't bother disabling it).

freekmurze:
I'd accept a PR that handles this.

SunMar:
Sure, created #287 :).

PR #287, opened 2020-07-13
IgnitionServiceProvider only setup queues when queue is available

SunMar:
Only setup the queues if we actually have queues available :) (see #272).


freekmurze:
Thanks!

PR #292, opened 2020-07-14
Add psalm check

freekmurze:


.github/workflows/psalm.yml
```diff
@@ -0,0 +1,33 @@
...
+        paths:
+            - '**.php'
+            - 'psalm.xml'
+
+jobs:
+    psalm:
+        name: psalm
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v1
```
riasvdv, line 14: There's a `actions/checkout@v2` that's a bit faster

src/Commands/TestCommand.php
```diff
@@ -72,7 +73,7 @@ protected function sendTestException()
         $testException = new Exception('This is an exception to test if the integration with Flare works.');
 
         try {
-            app('flare.client')->sendTestReport($testException);
```
mr-feek, line 75: hey @freekmurze was this breaking for you? I'm not sure I've added  support for the named aliases yet in laravels container. If so, I'll file that as a bug over at https://github.com/psalm/psalm-plugin-laravel and get working on it

freekmurze: Yeah, psalm didn't recognise the alias, but I was fine using the class name as well, so all good!

src/Http/Requests/ExecuteSolutionRequest.php
```diff
@@ -24,6 +24,7 @@ public function getSolution(): Solution
 
         abort_if(is_null($solution), 404, 'Solution could not be found');
 
+        /** @var Solution */
```
muglug, line 27: @mr-feek I think this would be made redundant by adding a `@psalm-assert` annotation to Laravel's `abort_if` function

PR #309, opened 2020-09-04
add loading state on share button

AdrianMrn:
This PR adds a loading state for the ignition share button, and stops people from sharing the same error more than once: 

Kapture 2020-09-04 at 13 36 32


freekmurze:
Very nice thanks!

PR #316, opened 2020-10-01
set default IGNITION_THEME to "auto"

hebbet:
so it follows the os preference by the user


freekmurze:
@AlexVanderbist @willemvb  I think this might be a better default. What do you think?

freekmurze:
We'll change to this in the next major version.

PR #317, opened 2020-10-01
Address Missing Mix Manifest Error

iAmKevinMcKee:
I encountered this error and realized that the solution is almost always that the front end assets have not been built. I thought this would be a nice solution for Ignition. Here's a screenshot of how it looks:

image


freekmurze:
Thanks!

Issue #322, opened 2020-10-21
PHP 8 test failures


Synchro:
Hello, it's your PHP 8 test stalker here again...

I've been trying to get ignition tests running on PHP 8. I've now got them running (thanks to your merges), but now have test failures. The problem now is that I have no idea what they mean, but I'm sure you will!

You can see the test config changes I've made [in my fork](https://github.com/Synchro/ignition/runs/1289050953?check_suite_focus=true#step:6:1).

```
Runtime:       PHP 8.0.0-dev
Configuration: /home/runner/work/ignition/ignition/phpunit.xml.dist
...........................................................F..... 65 / 88 ( 73%)
......F................                                           88 / 88 (100%)

Time: 00:01.327, Memory: 38.00 MB

There were 2 failures:

1) Facade\Ignition\Tests\Solutions\MergeConflictSolutionProviderTest::it_can_solve_merge_conflict_exception17
Failed asserting that false is true.

/home/runner/work/ignition/ignition/tests/Solutions/MergeConflictSolutionProviderTest.php:30

2) Facade\Ignition\Tests\Solutions\UndefinedVariableSolutionProviderTest::it_can_recommend_fixing_a_variable_name_typo
Failed asserting that false is true.

/home/runner/work/ignition/ignition/tests/Solutions/UndefinedVariableSolutionProviderTest.php:50
```

freekmurze:
Hey @Synchro, ❤️ your stalking 🙂 

Could you open up a PR with your changes. My colleague @AlexVanderbist will look into how to fix the tests.

AlexVanderbist:
Hi @Synchro, thanks for your work on this! Let's continue this discussion in your PR. Thanks!

Synchro:
OK, well you can push to my PR branch. I don't actually have any idea how to use ignition intentionally so I don't know that I'll be of any further help!

PR #323, opened 2020-10-22
PHP 8 test failures

Synchro:
Further to #322, I've created a fork which runs the test suite against PHP 8 by disabling platform checks in composer 2. This generally works, but as you can see it leads to 2 test failures which I don't understand.


Synchro:
Note that the failures only happen under PHP 8 – the tests run ok under 7.4.

AlexVanderbist:
Thanks! I've updated the tests to work with PHP 8 in this PR. As soon as everything reports green I'll merge it in :)

PR #332, opened 2020-11-10
Fix grammar & spelling in solution card component

unlikenesses:
Embarrassingly small, but it was nagging at me...


freekmurze:
Thanks!

Issue #335, opened 2020-11-22
Memory leak when using DB::insert


sfilzek:
Using DB::insert() with multiple records by passing an array of arrays (as per: https://laravel.com/docs/8.x/queries#inserts) seems to leak memory when ignition is installed.

(For reference, first opened on laravel/framework https://github.com/laravel/framework/issues/35319)

Ignition 2.5.2
Laravel 8.15.0

In this test I am inserting 50k records. If I perform 50k individual inserts then there is no leak. However if I inserts 10 batches of 5k records the leak appears. These tests were run by calling the functions directly from tinker:

````php
    public static function runBatch()
    {   
        $tempTable = 'csv_test_batch';

        Schema::create($tempTable, function (Blueprint $table) {
            $table->string('field_1');
            $table->string('field_2');
            $table->string('field_3');
            $table->string('field_4');
            $table->string('field_5');
        });
        
        $recordCount = 0;
        while ($recordCount < 50000) {
            $chunkSize = 5000;
            $data = [];
            while ($chunkSize-- > 0) {
                $data[] = [
                    'field_1' => 'Some data 1',
                    'field_2' => 'Some data 2',
                    'field_3' => 'Some data 3',
                    'field_4' => 'Some data 4',
                    'field_5' => 'Some data 5',
                ];
                $recordCount++;
            }
            DB::table($tempTable)->insert($data);
            info("Batch Insert: {$recordCount} records, memory: ". number_format(memory_get_usage()));
        }
    }
````

Running this function from tinker shows the leak:

````
[2020-11-22 07:44:11] local.INFO: Batch Insert: 5000 records, memory: 30,458,104  
[2020-11-22 07:44:12] local.INFO: Batch Insert: 10000 records, memory: 31,826,240  
[2020-11-22 07:44:12] local.INFO: Batch Insert: 15000 records, memory: 32,965,304  
[2020-11-22 07:44:12] local.INFO: Batch Insert: 20000 records, memory: 34,104,368  
[2020-11-22 07:44:12] local.INFO: Batch Insert: 25000 records, memory: 35,243,432  
[2020-11-22 07:44:12] local.INFO: Batch Insert: 30000 records, memory: 36,382,496  
[2020-11-22 07:44:12] local.INFO: Batch Insert: 35000 records, memory: 37,521,560  
[2020-11-22 07:44:12] local.INFO: Batch Insert: 40000 records, memory: 38,660,944  
[...]

Issue #307, opened 2020-09-03
Memory leak on providers registration


rennokki:
I came here to maybe start a discussion about a possible memory leak coming from this package.

The discussion is here: https://github.com/laravel/framework/issues/33952

The following comment indicates there is an issue with some of the recorders: https://github.com/laravel/framework/issues/33952#issuecomment-685172161

X-Coder264:
I just took a quick look so I might be wrong but the recorders `reset` method is called only when a worker tries to fetch a job from the queue which means there is no resetting with CLI long running commands so the arrays in question just keep getting bigger.

https://github.com/facade/ignition/blob/2.3.6/src/IgnitionServiceProvider.php#L432-L437

rennokki:
Does it help on calling the `reset()` method continuously?

mpskovvang:
I'm experiencing memory issues as well with queues.

Since day one I've `reporting.report_queries` set to `false`, but apparently this isn't respected since commit 9cc51f6b4e92703c331f1422d6c0ca0438a58206.

```
$this->app->make(QueryRecorder::class)->register();
$this->app->make(LogRecorder::class)->register();
$this->app->make(DumpRecorder::class)->register();
```
https://github.com/facade/ignition/commit/9cc51f6b4e92703c331f1422d6c0ca0438a58206#diff-cb4c5957c9c4ed5216b75f1bb182986fR81-R83

In the `register` method the `QueryRecorder` is only registered if enabled:

https://github.com/facade/ignition/blob/9cc51f6b4e92703c331f1422d6c0ca0438a58206/src/IgnitionServiceProvider.php#L101-L103

paras-malhotra:
I've submitted a PR for this.

freekmurze:
We'll continue the conversation on the PR.

rennokki:
Thanks for the follow-up guys!

Issue #302, opened 2020-08-07
Allowed memory size of 1073741824 bytes exhausted (tried to allocate 33554440 bytes)


flysion:
https://github.com/facade/ignition/blob/58dc4a8df3d41174c677a370b6779b694e4539e7/src/LogRecorder/LogRecorder.php#L35

When the code is running in console mode, this will cause the memory to continue to increase

kktsvetkov:
Can put some optional limit to the number of messages stored in `LogRecorder::$logMessages`, so when they are over the limit, the older messages are being discarded. However, I am not entirely sure how the garbage collection works and if the discarded objects will free their allocated memory

TheLevti:
Run into the same issue. There is currently no way to avoid the daemon to crash because at some point memory is full. I use for example Senty for exception reporting and their sdk also just tracks the last 100 events (logs) for example. It would be a good idea to limit this array to a maximum slots.

TheLevti:
Duplicate of #301

TheLevti:
This needs another fix. See conversation in #344 

#354 should fix this bug.

rubenvanassche:
This should be fixed with version 2.5.12, thanks @TheLevti for the fix.

andybobnewton:
this doesn't seem to have done enough. the array has stopped increasing in size but the memory usage still goes up from just logging a string.

just running the below will increase memory usage until fatal error (about 6 seconds on machine). 

`
    while (true) {
        Log::debug("got message" );
    }
`

Issue #301, opened 2020-08-06
Facade\Ignition\LogRecorder\LogRecorder::$logMessages memory limit


flysion:
//in vendor/facade/ignition/src/LogRecorder/LogRecorder.php

    public function record(MessageLogged $event): void
    {
        if ($this->shouldIgnore($event)) {
            return;
        }

        $this->logMessages[] = LogMessage::fromMessageLoggedEvent($event);
    }

Version:Laravel-7.22.4
Error:PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 4194312 bytes)

rubenvanassche:
This should be fixed with version 2.5.12, thanks @TheLevti for the fix!

PR #336, opened 2020-11-22
Don't register recorders and Flare utilities when Flare is not in use

paras-malhotra:
Fixes https://github.com/facade/ignition/issues/335, https://github.com/facade/ignition/issues/307, https://github.com/facade/ignition/issues/302, https://github.com/facade/ignition/issues/301, https://github.com/facade/ignition/issues/284, https://github.com/facade/ignition/issues/210, https://github.com/laravel/framework/issues/35319, https://github.com/laravel/framework/issues/35095, https://github.com/laravel/framework/issues/33952 and https://github.com/laravel/framework/issues/32054

### DIFF
--- a/src/IgnitionServiceProvider.php
+++ b/src/IgnitionServiceProvider.php
@@ -83,16 +83,32 @@ class IgnitionServiceProvider extends ServiceProvider
         $this
             ->registerViewEngines()
             ->registerHousekeepingRoutes()
-            ->registerLogHandler()
             ->registerCommands();
 
         if ($this->app->bound('queue')) {
             $this->setupQueue($this->app->get('queue'));
         }
 
-        $this->app->make(QueryRecorder::class)->register();
-        $this->app->make(LogRecorder::class)->register();
-        $this->app->make(DumpRecorder::class)->register();
+        if (config('flare.key')) {
+            $this->registerFlare()
+                ->registerLogRecorder()
+                ->registerDumpCollector();
+
+            if (config('flare.reporting.report_queries')) {
+                $this->registerQueryRecorder();
+            }
+
+            if (config('flare.reporting.anonymize_ips')) {
+                $this->app->get(Flare::class)->anonymizeIp();
+            }
+
+            $this->registerBuiltInMiddleware();
+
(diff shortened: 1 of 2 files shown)


src/IgnitionServiceProvider.php
```diff
@@ -83,16 +83,32 @@ public function boot()
...
             ->registerCommands();
 
         if ($this->app->bound('queue')) {
             $this->setupQueue($this->app->get('queue'));
         }
 
-        $this->app->make(QueryRecorder::class)->register();
-        $this->app->make(LogRecorder::class)->register();
-        $this->app->make(DumpRecorder::class)->register();
+        if (config('flare.key')) {
```
freekmurze, line 92: Using the Flare key to determine this is wrong. The collection information is display in Ignition as well.

I'm thinking it would be better to add keys in the `ignition` config file to determine which info should be collected. By default those new settings should be set to `true`. Tests should be added if the new settings are behaving as intended.

freekmurze (changes requested):

paras-malhotra:
> The collection information is display in Ignition as well.

It doesn't seem to be the case. Looking through the code, only Flare needs the Recorders. For example, a search for `getQueries` only gives one result in `AddQueries`, which is a middleware to the `Flare` class.

> By default those new settings should be set to true

If we set the default settings to true, the issues mentioned in this PR will not be solved. The main issue is that these recorders are causing a memory leak and should be disabled by default.

paras-malhotra:
On further review, I have found that the Ignition package indeed uses the query recordings (triggered by the Flare middleware). I don't think I have a detailed understanding of how ignition works and won't be able to invest much time in this PR.

I understand this is open source but sincerely hope that the maintainers put some energy into solving this. There have been multiple bug reports of memory leaks. Literally, for every issue reported to the Framework that pertains to memory leaks, @taylorotwell asks the OP to check with disabling Ignition first.

Also, as @driesvints pointed out this package is not only a dev dependency and is used for production as well. In the meantime, I have provided an alternative PR to the framework for this: https://github.com/laravel/framework/pull/35330.

PR #338, opened 2020-11-25
Fix double `$` on PHP 8

remzikocak:
On PHP 8.0 RC5 the variable name is shown with double $. This makes the button "Make variable optional" not working, because it searches for '{{ $$myVar }}' instead '{{ $myVar }}'.

This Pull request Strip's `$` sign from the beginning of the matched variable name. 

**Environment:**
OS: macOS BigSur 11.0.1
Laravel Valet Version: 2.13.2
PHP: 8.0 RC5

Tests are green on both PHP 7.4 and 8.0.

Bildschirmfoto 2020-11-25 um 23 20 53


freekmurze:
Thanks!

PR #340, opened 2020-12-04
Use Livewire compatible compiler engine when using Livewire 

austenc:
Applications using both Ignition and Livewire can be unruly at times due to Laravel expecting only a single view compiler engine to be bound in the container. 

In https://github.com/livewire/livewire/pull/2160 I have explained the issues in more detail. Due to Laravel expecting a single binding, it seems that this needs to be handled from both sides.

I am marking this as a draft until the final API is decided upon in the Livewire repo. Thank you!

### DIFF
--- a/src/IgnitionServiceProvider.php
+++ b/src/IgnitionServiceProvider.php
@@ -131,6 +131,10 @@ class IgnitionServiceProvider extends ServiceProvider
         });
 
         $this->app->make('view.engine.resolver')->register('blade', function () {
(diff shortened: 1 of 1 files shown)


calebporzio:
Just a heads up, I endorse this PR lol

freekmurze:
I'll review this as soon as it is marked as ready @austenc 👍

freekmurze:
Thank you!

austenc:
@freekmurze This should come out with the next Livewire release.

PR #343, opened 2020-12-28
Added support for Nova text editor

ngurney:
Added support for the Nova text editor ([nova.app](nova.app)) and updated the config comments to include it in the list of options.

_Disclaimer: I am new to programming and open source, if I missed something please reject_

### DIFF
--- a/config/ignition.php
+++ b/config/ignition.php
@@ -10,7 +10,7 @@ return [
     | Choose your preferred editor to use when clicking any edit button.
     |
     | Supported: "phpstorm", "vscode", "vscode-insiders",
-    |            "sublime", "atom"
+    |            "sublime", "atom", "nova"
     |
     */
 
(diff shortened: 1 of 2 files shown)


freekmurze:
Thank you so much for this. 👍

Enjoy your first merged contribution. 🎉

Please do it again in the future 🙂

ngurney:
Thanks @freekmurze! Hoping you can help me understand why it isn't working? I see the updates in my codebase in the config/ignition.php file, but I don't see them in the minified JS and I am not getting a working link. I ran `composer update` but no luck with that.

freekmurze:
I forgot to add the compiled js to the release.

Just corrected that in 2.5.5

ngurney:
@freekmurze cool, I can see that the js is updated now. I now get no content on the page though when I expect to see an error 😬 all I can see is the error title in my browser tab. Occurring on all browsers ever since I updated to 2.5.5, perhaps caused by something here?

freekmurze:
I could reproduce this, and @AdrianMrn is investigating this now.

freekmurze:
Seems like I accidentally merged in a breaking change in one of the js dependency. @AdrianMrn fixed it in the latest version.

PR #344, opened 2021-01-17
facade/ignition#302: Fix memory leaks caused by log and query recorder

### LINKED ISSUE #301: Facade\Ignition\LogRecorder\LogRecorder::$logMessages memory limit

### LINKED ISSUE #302: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 33554440 bytes)

TheLevti:
- Add optional log recorder limit to prevent memory leaks. Especially
  noticeable during long running tasks or daemons.
- Rewrote query recorder to not tightly couple it with the framework
  container when deciding whether to include bindings or limit the
maximum query count that is recorded.
- Added a config flag to disable log recording all together.
- Made sure that query and log recorders are not used when their enable
  flag is disabled.

Closes #302

Closes #301

### DIFF
--- a/src/IgnitionServiceProvider.php
+++ b/src/IgnitionServiceProvider.php
@@ -90,8 +90,14 @@ class IgnitionServiceProvider extends ServiceProvider
             $this->setupQueue($this->app->get('queue'));
         }
 
-        $this->app->make(QueryRecorder::class)->register();
-        $this->app->make(LogRecorder::class)->register();
+        if (config('flare.reporting.report_logs')) {
+            $this->app->make(LogRecorder::class)->register();
+        }
+
+        if (config('flare.reporting.report_queries')) {
+            $this->app->make(QueryRecorder::class)->register();
+        }
+
         $this->app->make(DumpRecorder::class)->register();
     }
 
@@ -106,8 +112,11 @@ class IgnitionServiceProvider extends ServiceProvider
             ->registerWhoopsHandler()
             ->registerIgnitionConfig()
             ->registerFlare()
-            ->registerLogRecorder()
             ->registerDumpCollector();
+
+        if (config('flare.reporting.report_logs')) {
+            $this->registerLogRecorder();
+        }
 
         if (config('flare.reporting.report_queries')) {
             $this->registerQueryRecorder();
@@ -279,13 +288,14 @@ class IgnitionServiceProvider extends ServiceProvider
         return $logLevel;
     }
 
-    protected function registerLogRecorder()
+    protected function registerLogRecorder(): self
     {
-        $logCollector = $this->app->make(LogRecorder::class);
-
-        $this->app->singleton(LogRecorder::class);
-
-        $this->app->instance(LogRecorder::class, $logCollector);
+        $this->app->singleton(LogRecorder::class, function (Application $app): LogRecorder {
+            return new LogRecorder(
+                $app,
+                $app->get('config')->get('flare.reporting.maximum_number_of_collected_logs')
+            );
+        });
 
         return $this;
     }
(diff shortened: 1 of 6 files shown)


freekmurze:
On first sight, this seems ok to me. 

@AlexVanderbist can you review and merge this?

CHANGELOG.md
```diff
@@ -400,3 +418,13 @@ All notable changes to `ignition` will be documented in this file
 ## 1.0.0 - 2019-08-30
 
 -   initial release
+
```
freekmurze, line 421: We don't want to maintain these links, could you remove them?

freekmurze (changes requested):

src/IgnitionServiceProvider.php
```diff
@@ -95,8 +95,12 @@ public function boot()
             $this->setupQueue($this->app->get('queue'));
         }
 
-        $this->app->make(QueryRecorder::class)->register();
-        $this->app->make(LogRecorder::class)->register();
+        if (config('flare.reporting.report_logs')) {
+            $this->app->make(LogRecorder::class)->register();
+        }
+        if (config('flare.reporting.report_queries')) {
```
rubenvanassche, line 101: Could you add some room between these statements? It feels a  bit crammed up here.

src/IgnitionServiceProvider.php
```diff
@@ -324,30 +332,38 @@ protected function registerCommands()
...
-        ->map(function (string $middlewareClass) {
-            return $this->app->make($middlewareClass);
-        });
+        ];
+        if (config('flare.reporting.report_logs')) {
+            $middlewares[] = AddLogs::class;
+        }
+        $middlewares[] = AddDumps::class;
+        if (config('flare.reporting.report_queries')) {
+            $middlewares[] = AddQueries::class;
```
rubenvanassche, line 359: Same here, some room between these statements will make them more readable

CHANGELOG.md
```diff
@@ -1,24 +1,42 @@
...
 
 - fix double `$` on PHP 8 (#338)
 
-## 2.5.7 - 2020-12-29
+## [2.5.7] - 2020-12-29
 
 - fix for breaking change in highlight.js (fixes 2.5.5)
 
-## 2.5.6 - 2020-12-29
+## [2.5.6] - 2020-12-29
```
rubenvanassche, line 39: Could you change the changelog 😄 to the format is was before? Like this:

```
## VERSION - DATE
-THING THAT CHANGED
```

Otherwise we lose consistency between our packages

rubenvanassche (changes requested):
Some minor changes request, but a great PR, look forward to merge it!

AlexVanderbist (approved):
Apart from the minor remarks by my colleagues; LGTM!

CHANGELOG.md
TheLevti: Sure, was just a habit and suggestion to use a standardized format.

src/IgnitionServiceProvider.php
TheLevti: done

src/IgnitionServiceProvider.php
TheLevti: done

CHANGELOG.md
TheLevti: removed changelog changes

rubenvanassche:
Thanks, looks great!

rimace:
This change is causing a memory leak for me.
I am importing ~300,000 models. First 1,000 models are read and bundled into one transaction, then the query is fired and then next 1,000 and so on.
Usually my import uses always around 25 MB RAM, but since version 2.5.11 my importer reaches 500 MB RAM after a while.
Will report a proper issue with an example later today.

TheLevti:
> This change is causing a memory leak for me.
> I am importing ~300,000 models. First 1,000 models are read and bundled into one transaction, then the query is fired and then next 1,000 and so on.
> Usually my import uses always around 25 MB RAM, but since version 2.5.11 my importer reaches 500 MB RAM after a while.
> Will report a proper issue with an example later today.

Thanks for the feedback. Please let me know the details. For my use case the memory leak that I had previously was fixed once the array is topping out at the configured limit.

rimace (commented):
That's it.
The QueryRecorder uses a new config which didn't exist in my config cache (forgot to recreate it).
Because QueryRecorder was activated but the maximum_number_of_collected_logs didn't exist in my config cache yet, it was null which deactivated the "maximum logs" of 200 (by default).
So it was my fault for not recreating the config cache. Still I think the name of the config variable is wrong there.

src/IgnitionServiceProvider.php
```diff
@@ -319,30 +329,42 @@ protected function registerCommands()
...
-        $queryCollector = $this->app->make(QueryRecorder::class);
-
-        $this->app->singleton(QueryRecorder::class);
-
-        $this->app->instance(QueryRecorder::class, $queryCollector);
+        $this->app->singleton(QueryRecorder::class, function (Application $app): QueryRecorder {
+            return new QueryRecorder(
+                $app,
+                $app->get('config')->get('flare.reporting.report_query_bindings'),
+                $app->get('config')->get('flare.reporting.maximum_number_of_collected_logs')
```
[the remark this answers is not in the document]
TheLevti: Oh jeez, this is the classic copy pasta mistake. Will provide a fix for it.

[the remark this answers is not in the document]
TheLevti: https://github.com/facade/ignition/pull/354

PR #354, opened 2021-02-14
facade/ignition#344: Fix wrong config usage

### LINKED ISSUE #301: Facade\Ignition\LogRecorder\LogRecorder::$logMessages memory limit

### LINKED ISSUE #302: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 33554440 bytes)

TheLevti:
This fixes a copy paste bug that got overseen. A config value has not been used at all, instead a different config value has been used in the wrong place.

Closes #302 

Closes #301

### DIFF
--- a/src/IgnitionServiceProvider.php
+++ b/src/IgnitionServiceProvider.php
@@ -340,7 +340,7 @@ class IgnitionServiceProvider extends ServiceProvider
             return new QueryRecorder(
                 $app,
                 $app->get('config')->get('flare.reporting.report_query_bindings'),
-                $app->get('config')->get('flare.reporting.maximum_number_of_collected_logs')
+                $app->get('config')->get('flare.reporting.maximum_number_of_collected_queries')
             );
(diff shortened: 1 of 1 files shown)


rubenvanassche:
Thanks for the fix!

PR #334, opened 2020-11-16
Fix MakeViewVariableOptionalSolution to disallow stream wrappers and files that do not end in .blade.php

cfreal:
Hello, as discussed by email, this fixes a serious vulnerability.
Hopefully my code is OK-ish.

PR #356, opened 2021-02-18
Fix MakeViewVariableOptionalSolution to disallow stream wrappers and files that do not end in .blade.php

anasmirza534:
This is already fixed in 2.5.2, See https://github.com/facade/ignition/pull/334

I could not update to 2.5.2 due to some dependent package required php 7.3, currently clients site is running in php 7.2

On branch 2.4.1-branch
Changes to be committed:
	modified:   src/Solutions/MakeViewVariableOptionalSolution.php


AlexVanderbist:
Thanks! Tagged as 2.4.2

PR #364, opened 2021-03-30
Disable executing solutions on non-local environments or from non-local IP addresses

AlexVanderbist:
Title says it all! Includes tests.


freekmurze:
Thanks!

PR #366, opened 2021-03-30
Adds a debug warning when having debug enabled on a non-local environment

riasvdv:
Adds the following warning:

image

With a link to https://flareapp.io/docs/ignition-for-laravel/security for more information (page still has to be added).

When APP_DEBUG is `true` and the app's environment is anything other than `local`


src/ErrorPage/ErrorPageViewModel.php
```diff
@@ -34,6 +34,12 @@ class ErrorPageViewModel implements Arrayable
     /** @var array */
     protected $defaultTabProps = [];
 
+    /** @var string */
+    protected $app_env;
```
freekmurze, line 38: Shouldn't we camelCase these?

freekmurze (changes requested):

src/ErrorPage/ErrorPageViewModel.php
riasvdv: Good point, fixed!

freekmurze:
Thanks!

PR #372, opened 2021-04-07
Censor request body

freekmurze:
This PR will censor all password fields in a request by default.

The fields to be censored can be modified in the config file.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "php": "^7.2.5|^8.0",
         "ext-json": "*",
         "ext-mbstring": "*",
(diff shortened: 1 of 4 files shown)


AlexVanderbist (approved):

composer.json
```diff
@@ -13,7 +13,7 @@
         "php": "^7.2.5|^8.0",
         "ext-json": "*",
         "ext-mbstring": "*",
-        "facade/flare-client-php": "^1.3.7",
+        "facade/flare-client-php": "dev-censor-request-body as 1.3.7",
```
rubenvanassche, line 16: This should be reverted back again!

freekmurze: Good call!

Issue #375, opened 2021-04-08
Argument 1 passed to Facade\FlareClient\Flare::censorRequestBodyFields() must be of the type array, null given, called in /var/www/vendor/facade/ignition/src/IgnitionServiceProvider.php on line 134


tarreislam:
Hi, the new update breaks everything because of array type hinting on null value in `register` part

`Argument 1 passed to Facade\FlareClient\Flare::censorRequestBodyFields() must be of the type array, null given, called in /var/www/vendor/facade/ignition/src/IgnitionServiceProvider.php on line 134`

Solved by manually adding  missing properties to `config\flare.php`  

` 'censor_request_body_fields' => ['password'],`

tarreislam:
created pull

https://github.com/facade/ignition/pull/376

rubenvanassche:
Thanks for bringing this to our attention, we've just released ignition 2.8.1 which should fix the issue.

PR #376, opened 2021-04-08
Update IgnitionServiceProvider.php

tarreislam:
censorRequestBodyFields type hints array, but null is given if config is old.


tarreislam:
https://github.com/facade/ignition/issues/375

rubenvanassche:
Thanks!

PR #381, opened 2021-04-29
Avoid making call to flare

freekmurze:
When using the Flare driver, without an API key a http request to Flare is made.
We should not do that.

Fixes https://github.com/facade/flare-client-php/issues/22

### DIFF
--- a/src/IgnitionServiceProvider.php
+++ b/src/IgnitionServiceProvider.php
@@ -258,10 +258,6 @@ class IgnitionServiceProvider extends ServiceProvider
 
     protected function registerLogHandler()
     {
-        if (! config('flare.key')) {
-            return $this;
-        }
-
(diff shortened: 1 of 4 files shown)


AlexVanderbist (approved):
🚀

AlexVanderbist (changes requested):
Maybe we can move the check for having an API key to this method? https://github.com/facade/ignition/blob/master/src/Logger/FlareHandler.php#L71

Seems more appropriate but up to you :)

tpetry:
> Maybe we can move the check for having an API key to this method? https://github.com/facade/ignition/blob/master/src/Logger/FlareHandler.php#L71
> 
> Seems more appropriate but up to you :)

When debugging the issue i had the same idea 😅

freekmurze:
@AlexVanderbist that's better, I'll do it that way, thanks!

PR #390, opened 2021-05-25
(fix): use application contract

edwinsiebel:
Instead of using the concrete `Application` implementation, uses the contract/interface `Application`.


freekmurze:
Tests seem to be failing, so passing on this one.

PR #398, opened 2021-07-12
Fix sending queued reports on Laravel Vapor queues

AlexVanderbist:
Laravel Vapor processes queued jobs one by one without firing the `Looping` event in between jobs. Vapor also doesn't call PHP's registered shutdown functions (on the queue lambda at least). Exceptions thrown in Vapor jobs will not get reported because of these two issues.

The solution is to use both the queue `before` and `after` events. `Before` will send any pending reports (from a before the queued job) and make sure the dump/log/query recorders are emptied before executing the queued job. `After` will report any exceptions thrown during job execution.

Because we're using both `before` and `after`, there's some overlap but as this only causes some internal arrays to be emptied twice, this can be ignored.


rodrigopedra:
@AlexVanderbist this seems to have caused #401 

Had to revert to 2.11.0 due to errors on horizon

PR #402, opened 2021-07-20
Make resetFlare public for event listeners

### LINKED ISSUE #401: Call to protected method 'resetFlare'
With 2.11.1, there seems to be an issue with 'dispatch_now' jobs. We're getting the following errors directly after deploying the new version:

`Call to protected method Facade\Ignition\IgnitionServiceProvider::resetFlare() from context 'Illuminate\Events\Dispatcher'`

dmason30:
Fixes #401 

Alternative fix is in #403


AlexVanderbist:
We just merged #403 but thanks for contributing a fix so quickly!

PR #404, opened 2021-07-23
Disable executing solutions on non-local environments or from non-local IP addresses on version 1.x

Parsk:
Ported code from commit https://github.com/facade/ignition/pull/364 by AlexVanderbist to branch v1, with tests. Matter also discussed here https://github.com/facade/ignition/issues/350#issuecomment-884710367

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -11,7 +11,6 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
-            fail-fast: false
             matrix:
                 php: [8.0, 7.4, 7.3, 7.2]
                 laravel: [6.*, 5.8.*, 5.7.*, 5.6.*, 5.5.*]
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -11,6 +11,7 @@ jobs:
         runs-on: ${{ matrix.os }}
 
         strategy:
+            fail-fast: false
             matrix:
                 php: [8.0, 7.4, 7.3, 7.2]
                 laravel: [6.*, 5.8.*, 5.7.*, 5.6.*, 5.5.*]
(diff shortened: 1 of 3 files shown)


freekmurze:
Seems like the tests are failing for this one. Could you take a look at that?

Parsk:
I would seem that  `        if ($this->app->runningInConsole()) {
            return $this;
        }` here https://github.com/facade/ignition/blob/v1/src/IgnitionServiceProvider.php#L132 seems to disable Laravel builds from registering the routes? PHPUnit runs successfully on a stand alone deployment

freekmurze:
I'm thinking it breaks because of changes in this PR. Tests seem to be running fine on the `master` branch.

Parsk:
Actually just noticed that the P7.2 - L5.8.* had successfully ran all tests, while P7.2 - L5.6.* had failed. So I really don't see a reason why it would fail. I downgraded and tested with PHPUnit 7.5.20  just to be sure, and it was all OK. Also the test code and the registered paths are identical with the current master.

Is there a possibility to run all the tests again without it stopping on first failure?

freekmurze:
You'll have to set `fail-fast` to `false` in [the run-tests workflow](https://github.com/facade/ignition/blob/dc49305538aeb77e4c89eba57cee4ceff9e42f33/.github/workflows/run-tests.yml).

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

Parsk:
Alright, I tracked the issue down. Basically the environment setting `APP_RUNNING_IN_CONSOLE` was only introduced in Laravel v5.7.14, so the function `runningInConsole()` only looks at the `php_sapi_name` in older versions. So the in old versions the routes basically don't get registered.

Is it OK to skip these tests below 5.7.14? And my apologizes for wasting your time for this small change.

freekmurze:
> Is it OK to skip these tests below 5.7.14? 
Ok, make it so! 🙂 

> And my apologizes for wasting your time for this small change.

No worries

freekmurze:
Seems like the tests are still failing for this one 😬

Parsk:
Seems like there is some issues with the runners, those that have failed have the error `"This request was automatically failed because there were no enabled runners online to process the request for more than 1 days."`

Shall I push an empty commit to run the tests again?

freekmurze:
All good now, thanks!

freekmurze:
✅ functionality already available in spatie/laravel-ignition

PR #406, opened 2021-08-05
Resolve view engine when needed

foremtehan:
in my laravel app (telegram) i dont have any View system so, when i disable the `Illuminate\View\ViewServiceProvider::class` in app.php i cant use Cache system and i'll get: `Target class [view.engine.resolver] does not exist.` error, so:


freekmurze:
It seems this breaks the tests. Could you take a look at that?

foremtehan:
hey i decided to remove this package as dev-dependency sorry

PR #408, opened 2021-08-12
Fix Circular references in Route parameters #352

### LINKED ISSUE #352: Circular references in Route parameters
Hi there,

We found an issue with circular references when Ignition is rendering an Exception page. In our case there is a model containing a reference to itself in its relations that has been set as a route parameter. Whenever ignition is displaying an error page, it breaks and throws an nginx 502 bad gateway. This issue is resolved by removing the circular reference in the object stored as a route parameter. 

My guess is that the "Route parameters" preview under the "App" tab does not handle circular references, which could actually be quite common...

Here is a simplified screenshot of what we cannot display correctly :

image

NEGits:
This fixes #352, by unloading Model relations before casting Models to array. 

I'm not familiar enough with testing Route Model binding to add a test for this unfortunately, hope that's ok.

### DIFF
--- a/src/Context/LaravelRequestContext.php
+++ b/src/Context/LaravelRequestContext.php
@@ -3,6 +3,7 @@
 namespace Facade\Ignition\Context;
 
 use Facade\FlareClient\Context\RequestContext;
+use Illuminate\Database\Eloquent\Model;
 use Illuminate\Http\Request;
 use Throwable;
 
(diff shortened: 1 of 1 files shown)


AlexVanderbist:
Thanks, looks good!

freekmurze:
✅ ported to spatie/laravel-ignition

PR #411, opened 2021-08-17
Use npm ci instead of install

jonassiewertsen:
I would suggest to use `npm ci` instead of `npm install`. 

`npm ci` does only install dependencies as locked in the package.lock file. This does mean, that `npm ci` can be compared with `composer install` in PHP. 

**Why not to use `install`**
Install might update dependencies and might create a new package.lock. 

**Downsides of `install`**
In case someone does not have a package.lock file, npm will throw an error message saying that you need to run `npm install`. 

This may sound like a downside, but personally I think this is fine as the same is required if using `composer install` in PHP.


freekmurze:
Thanks!

freekmurze:
✅ Ported to spatie/laravel-ignition

PR #417, opened 2021-09-09
Add support for VS Codium editor

HajMo:
This PR will add support for ["VS Codium"](https://vscodium.com) which is a release from the "VS Code" but without  MS branding/telemetry/licensing.


freekmurze:
Thanks!

freekmurze:
@AdrianMrn could you port this change to the JS we'll use on spatie/ignition?

PR #422, opened 2021-10-04
Enable PHP 8.1 builds

driesvints:
Enables PHP 8.1 builds. No necessary changes it seems based from the current test suite. 

I've removed the testbench toggling in the CI setup as the composer can resolve the correct one. I've also removed the separate PHP 8 >= vs PHP 8 < composer dependency requirements as `--ignore-platform-req=php` is no longer required at this time in PHP 8.1's development cycle.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -40,7 +40,7 @@ jobs:
 
             - name: Install dependencies
               run: |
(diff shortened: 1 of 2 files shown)


freekmurze:
Nice, thanks!

freekmurze:
We would welcome these changes at spatie/ignition and spatie/laravel-ignition as well.

PR #429, opened 2021-11-26
Fix solution overflow

faissaloux:
Hello 👋🏻
In this PR I've fixed this overflow problem.

# Before
Screenshot from 2021-11-26 20-22-59

# After
Screenshot from 2021-11-26 20-36-27


willemvb:
Hey @faissaloux , thanks for pointing this out! 
I've created an alternative solution in #431 that scrolls the entire card (just like the exception card above) with a fancy scrollbar. 
It also preserves the `max-width` on the `solution-content` to prevent too long text-lines under normal circumstances.

PR #431, opened 2021-11-29
Allow overflow-x on solutions with unbreakable words

willemvb:
Alternative solution for https://github.com/facade/ignition/pull/429

- Allow overflow on the entire solutions card cfr. to the Exception Card
- Use smaller scrollbar

image


freekmurze:
Thanks!

PR #433, opened 2021-11-29
fix solutions section padding

faissaloux:
Hello :wave:
In this PR I have fixed solutions section padding.

# Before
Screenshot from 2021-11-29 16-29-22

# After
Screenshot from 2021-11-29 17-31-23


resources/js/components/Solutions/SolutionCard.vue
```diff
@@ -51,7 +51,10 @@
                         ></div>
```
willemvb, line 51: Shouldn't `solution-description` be used here as well?

freekmurze:
Thanks!

PR #434, opened 2021-11-29
Cross-platform PHPStorm editor URL

bleuscyther:
I used the JetBrain toolbox extension + some extra documentation to generate the correct URL for the new IDE with Toolbox. 

This should work on all platforms.  

Some points of concern

- Compatibility with old PhpStorm versions, especially before the toolbox app
- Seem to open file one line after intended 

fixes this issue: https://github.com/facade/ignition/pull/112


bleuscyther:
Some more updates on this issue. 

Standalone PhpStorm without Toolbox will have no issue with the phpstorm://  only when managed by Toolbox.
https://youtrack.jetbrains.com/issue/IDEA-283631
This pull request may break the experience for some people...

At this point maybe the user should be more involved in choosing the IDE, or there should be a dropdown in the UI about that where they can toggle between IDEs.

- Auto
- JetBrain PhpStorm
- JetBrain Toolbox + PhpStrom
- ...
😅

Dunno, that would be one solution where packages like Laravel won't need to add extra stuff for the user

bleuscyther:
⚠This also breaks when the `.idea` directory is not present.

freekmurze:
We're going to leave things as they are for now.

nicolus:
@bleuscyther

> Standalone PhpStorm without Toolbox will have no issue with the phpstorm:// only when managed by Toolbox. 

On windows 10 with my projects in WSL2 the `phpstorm://` protocol has **never** worked for me. The Toolbox on the other hand works flawlessly with the `jetbrains://` protocol.

It seems to me the correct way would be to have 2 settings as you said : "PhpStorm" and "Jetbrains Toolbox", and also provide an option to enter the project name manually, as I feel like trying to autodetect the project name from files is bound to fail for someone at some point.

PR #457, opened 2022-06-30
notice if dots have been used in the view name

WebPajooh:
> View directory names should not contain the . character.
https://laravel.com/docs/9.x/views#nested-view-directories

Today I got this message because I had used a name like `foo.bar.blade.php` for my view file:
>View [foo.bar] not found.
Did you mean foo.bar?  

It's funny and confusing, not clear and helping!

This happens because Laravel replaces the dots with slashes so it thinks that we want this file:
```
views
    foo
        bar.{extension}
```

The solution description is pretty deceptive, so I have fixed that by adding a condition that you can see it in my commit. This is the new solution description when anyone makes that mistake:
> View names should not contain the . character.


freekmurze:
Thanks!

PR #458, opened 2022-07-08
add php 8.1 support

aharen:


aharen:
add PHP 8.1 support

freekmurze:
We already support PHP 8.1

PR #466, opened 2022-12-12
Replace ${var} with {$var}

Kunimaru:
```
Uncaught Whoops\Exception\ErrorException: Using ${var} in strings is deprecated, use {$var} instead in /home/user/app/vendor/facade/ignition/src/SolutionProviders/MergeConflictSolutionProvider.php:52
```

In php8.2 I got the previous error and it was solved by this commit.
I propose you guys to apply this change to prepare for php8.2 supoprt in the future.


rubenvanassche:
Please use spatie/laravel-ignition, this package isn't maintained anymore.

PR #469, opened 2023-01-23
Update psalm and add it to CI

lptn:
psalm/plugin-laravel has been recently updated and together with new Psalm versions can detected more issues. Due to other dependencies, it's not possible to use the latest one version of the plugin, but even [v2.0.2](https://github.com/psalm/psalm-plugin-laravel/releases/tag/v2.0.2) works a way better

### DIFF
--- a/.github/workflows/psalm.yml
+++ b/.github/workflows/psalm.yml
@@ -1,0 +1,56 @@
+name: Psalm
+
+on:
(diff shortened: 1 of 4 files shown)


freekmurze:
This package has been deprecated, the new version is [here](https://github.com/spatie/ignition)

